Issue #9, opened 2020-01-16
LVC conversion: Are error outputs needed?


moble:
The `errors` produced by romspline and stored in the output LVC format files do not seem useful for our purposes; they look like they're more relevant to studying the greedy algorithm itself.  Storing them increases the size of each LVC-format file by almost 50%.

Each time series in LVC format is currently stored as a group containing five datasets:
  1. `X`, representing the sample times (aka knots of the spline)
  2. `Y`, representing the data values at those times
  3. `deg`, giving the degree of the spline used in the greedy algorithm
  4. `tol`, giving the target tolerance allowed by the greedy algorithm
  5. `errors`

This last dataset represents the largest error that the greedy algorithm encountered at each step as it was progressively choosing more and more knots to use for the spline.  For example, the first element in any given `errors` dataset is the largest error when approximating the entire data series in question by a spline with just 6 knots; the second element gives the largest error with 7 knots.  I don't see any connection between these quantities and anything we actually care about with the final result, which will typically use hundreds to thousands of knots.  In particular, the points with those largest errors are then added to the set of knots, so that the error for the final result *at* those points should actually be 0.0 in all cases.

The `romspline.ReducedOrderSpline.write` function that we use to write these files has a `slim` option that, if set to `True` skips writing this `errors` dataset.  Can we do so?  Does anyone use those datasets?  This would reduce our LVC file sizes by almost a third, and might simplify certain other things that I think might need to happen.

moble:
On [a vacuum phone call](https://wiki.black-holes.org/vacuum_meetings/2020#january_22_2020), we agreed that this field is no longer needed.  I think as a matter of safety, I might actually keep this around, in the form of an array of size `(1,)` and just containing a copy of the tolerance.

moble:
Closed by #10

Issue #8, opened 2020-01-16
LVC conversion: better tolerances


moble:
@geoffrey4444 wants to convert our entire catalog to LVC format.  The main problem with that is that the conversion process takes of order hours per waveform, so without any improvements, we would expect this process to take of order a year on a single computer.

After looking through the data, it's become clear that the real problem here is the phase — more specifically, the use of phase as one of the data series while requiring the same tolerance on the amplitude and phase.  The result is that splines of the phase in the output LVC format
  1. take **hundreds to thousands of times longer** to construct than than they need to, and
  2. occupy **tens to hundreds of times more space** than they need to

for the resulting _waveform_ to have the desired accuracy.

Each mode of the input waveform is first decomposed into `amp` and (unwrapped) `phase`, and then passed to `romspline.ReducedOrderSpline`, which builds a spline that should be able to reproduce the input data to within an L-infinity tolerance of 1e-6.  Note that this tolerance is absolute, not relative.  

Presumably, the goal here is to produce a waveform for consumption by, e.g., LAL that is accurate to 1e-6.  This is basically achieved with `amp`.  The `amp` will typically be no larger than 0.1 (or orders of magnitude less for higher-order modes) except very near merger, meaning the relative tolerance is more like 1e-5 or less restrictive — but the absolute error will indeed be 1e-6.

Meanwhile the `phase` will generally be tens to tens of thousands (depending on junk and/or precession), meaning that the relative tolerance will be 1e-7 or more restrictive.  This is why the `phase` splines perform so much worse than `amp`.

The answer seems obvious: rather than requiring |δ`phase`|<1e-6, we need to be able to require `amp`&ast;|δ`phase`|<1e-6.  Better yet, to keep the results as accurate as before, we should require |δ`amp`|<5e-7 and `amp`&ast;|δ`phase`|<5e-7, because now the latter will be contributing roughly equally with the former and L-infinity errors should add linearly.

But then, would LVC people accept such waveforms?  I know everyone talks about how the phase really needs to be accurate, but I claim it will still be (at least nearly) as accurate as it was before.  Specifically, this is not a cumulative phase error; it is an interpolation error.  This means that there will still be many times throughout the output waveform that will have precisely zero phase error (relative to the input waveform), but there will be small fluctuations of the phase error between those points.  These enter into the waveform in very much the same way as the amplitude errors, so they should be considered equivalent.

moble:
[...]

Issue #7, opened 2020-01-15
LVC conversion: spline degree


moble:
Moving another part of sxs-collaboration/catalog_tools#3 to this repo, because the code has moved.

@SergeiOssokine said
> Currently when we call romspline to generate the spline we use the default value for the degree of the spline, which is 5. However, in LALSuite, when the data is being reconstructed the interpolation is done using cubic splines. Should we also just use cubic splines?

@geoffrey4444 said
> I vote for using the same degree as before, not cubic splines, for consistency.

I vote for changing it to use cubic splines.  I think there are two very good reasons (and one maybe-good reason) for this:
  1. Cubic splines are significantly faster.  In one quick test on the ell<=3 modes of one waveform, it's more than twice as fast to use cubic splines.
  2. That same test actually shows a ~3% *reduction* in output file size.  This actually agrees with what I expected, because higher-degree splines tend to get unstable, so they effectively need more control points — especially when the data are noisy or otherwise not particularly smooth.  While the biggest modes are fairly smooth, most of the modes are not.  There's a rule of thumb that says cubic splines are typically the best choice for data that's appropriately sampled, and that seems to be true here.
  3. The tolerance doesn't mean what we think it means when comparing splines of different degrees.  If tolerance 1e-6 is achieved for the default `deg=5`, I see no reason to expect that to be true of a waveform reconstructed with `deg=3`.  This may seem to conflict with the previous point, because some data series have more points when using `deg=5`; but some data series also have fewer points, so there are surely some series that will not be interpolated properly.

It looks like points 1 and 2 are the same or stronger if I go to higher ell values, but I just don't have the time to wait around to completely answer that question.

I've added options to `sxs.format.lvc.convert_simulation` allowing the user to select the desired spline degree and tolerance.  They're currently set to the romspline defaults, but I'm suggesting changing the former to 3, rather than 5.

moble:
On [a vacuum phone call](https://wiki.black-holes.org/vacuum_meetings/2020#january_22_2020), we agreed that `deg=3` is the best choice going forward.

moble:
Closed by #10

PR #10, opened 2020-01-16
Generalize romspline's _greedy function to allow more general tolerances

moble:
This PR basically extracts the functionality that we actually use from romspline, rewrites it to conform with more current python, and generalizes it to accept a function in place of `tol` (so that we can normalize to amplitude when splining phase).  It also removes the `error` datasets from the LVC output.  Combining these two points provides us with a ~90x speedup, while producing files ~11x smaller, which still result in the same accuracy for LVC waveforms.  I've also changed the default spline degree from 3 to 5 to align with LVC usage, which provides a further speedup and slightly smaller files.

This addresses #7, #8, and #9.


moble:
I have not yet removed `romspline` entirely, because `horizons.py` uses it in a way that's slightly awkward for me; it'll just take a little more reorganization of the code in that file to finish this.

PR #32, opened 2020-10-13
Run tests on python 3.9

moble:
Once the python 3.9 ecosystem is running (specifically, when numpy scipy, and numba are available for all platforms), the github action should run on python 3.9 as well.  It's currently failing on this branch because they don't install correctly.  We should just re-run it every few days until it passes, then merge this.


moble:
Waiting on https://github.com/numba/numba/pull/6579

PR #52, opened 2022-03-14
Add PN generator.

dongzesun:
Add python code to generate PN waveform. There are still problems with TaylorT4 and T5.


moble:
I guess I should just quickly write my main comment, which is that most of the files you've added here should still just be in https://github.com/moble/PostNewtonian.  If you've made changes to any of the files that are already in that repo, you should just make a pull request on that repo.  I would guess that you just need one or two notebooks in that repo to generate the python code that actually creates the PN waveforms.

On the other hand, the code  that should go into this repo should be one or two files generated by all those notebooks and utilities in that repo, and some high-level function that you actually call to evolve a PN system.  It shouldn't require more than one or two new files in this repo — for example, `sxs/sxs/waveforms/pn.py`.

PR #54, opened 2022-03-23
Adding tag to json_data for memory correction

keefemitman:
Adding a tag to json_data for memory correction. But `sxs.waveforms.rpxmb.save()` is set to always raise an error?

### DIFF
--- a/sxs/waveforms/memory.py
+++ b/sxs/waveforms/memory.py
@@ -7,7 +7,7 @@ this:
 
 ```python
 h = sxs.load("SXS:BBH:0123/Lev/rhOverM", extrapolation_order=3)
-h_with_memory = sxs.waveforms.memory.add_memory(h, start_time=1000.0)
+h_with_memory = sxs.waveforms.memory.add_memory(h, integration_time=1000.0)
 ```
 
 """
@@ -139,7 +139,7 @@ def J_m(h, Psi2):
     return J_m
 
 
-def J_E(h, start_time=None):
+def J_E(h, integration_time=None):
     """Energy flux contribution to electric part of strain
 
     Calculated according to Eq. (17b) of 'Adding Gravitational Memory to Waveform
@@ -149,7 +149,7 @@ def J_E(h, start_time=None):
     ----------
     h : WaveformModes
         WaveformModes object corresponding to the strain
-    start_time : float, optional
+    integrations_time : float, optional
         The time at which the energy flux integral should begin.  The default is
         `h.t[0]`.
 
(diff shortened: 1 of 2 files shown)


sxs/waveforms/memory.py
```diff
@@ -149,7 +149,7 @@ def J_E(h, start_time=None):
     ----------
     h : WaveformModes
         WaveformModes object corresponding to the strain
-    start_time : float, optional
+    integrations_time : float, optional
```
markscheel, line 152: minor: integration_time, not integrations_time

moble:
Currently, this PR would produce JSON data that (ignoring checks to see if the attributes exist) essentially looks like
```python
            "transformations": {
                "truncation": L2norm_fractional_tolerance,
                "boost_velocity": w.boost_velocity.tolist(),
                "space_translation": w.space_translation.tolist()
            },
            "memory_correction": {
                "integration_time": w.memory_integration_start_time
            },
```
The `transformations` group is actually intended to record *all* the transformations — not just BMS transformations.  Maybe `modifications` would be a better name for that group.  Anyway, the memory correction info should probably go in there.

But this makes me think it might actually be better to be more explicit about these arguments, by including the name of the function doing the modification, and then its arguments as kwargs.  So it would look more like this:
```python
            "transformations": {
                "truncate": {"tol": L2norm_fractional_tolerance},
                "transform": {
                    "boost_velocity": w.boost_velocity.tolist(),
                    "space_translation": w.space_translation.tolist()
                },
                "add_memory": {"integration_start_time": w.memory_integration_start_time}
            },
```
(and maybe change `transformations` to `modifications` or something).  This could break any scripts that read `boost_velocity` and `space_translation` from the mini catalog, but I can't imagine there are enough such scripts to worry about.

Any opinions?

keefemitman:
> Currently, this PR would produce JSON data that (ignoring checks to see if the attributes exist) essentially looks like
> 
> ```python
>             "transformations": {
>                 "truncation": L2norm_fractional_tolerance,
>                 "boost_velocity": w.boost_velocity.tolist(),
>                 "space_translation": w.space_translation.tolist()
>             },
>             "memory_correction": {
>                 "integration_time": w.memory_integration_start_time
>             },
> ```
> 
> The `transformations` group is actually intended to record _all_ the transformations — not just BMS transformations. Maybe `modifications` would be a better name for that group. Anyway, the memory correction info should probably go in there.
> 
> But this makes me think it might actually be better to be more explicit about these arguments, by including the name of the function doing the modification, and then its arguments as kwargs. So it would look more like this:
> 
> ```python
>             "transformations": {
>                 "truncate": {"tol": L2norm_fractional_tolerance},
>                 "transform": {
>                     "boost_velocity": w.boost_velocity.tolist(),
>                     "space_translation": w.space_translation.tolist()
>                 },
>                 "add_memory": {"integration_start_time": w.memory_integration_start_time}
>             },
> ```
> 
> (and maybe change `transformations` to `modifications` or something). This could break any scripts that read `boost_velocity` and `space_translation` from the mini catalog, but I can't imagine there are enough such scripts to worry about.
> 
> Any opinions?

Yeah I like the idea of referencing the functions that do the modification, but then I think we should change `transformations` to `modifications`

PR #58, opened 2022-04-06
add_memory now optionally corrects psi4 as well as h.

markscheel:
The return value of add_memory is now a tuple.

(Instead, I could have made the return value of add_memory be a tuple if psi4 is not None but just a single value if psi4 is None.  Not sure if tuple-all-the-time is more or less confusing than return-value-depends-on-inputs, so I chose the former. If you want the latter, let me know).

### DIFF
--- a/sxs/waveforms/memory.py
+++ b/sxs/waveforms/memory.py
@@ -256,10 +256,10 @@ def add_memory(h, integration_start_time=None, psi4=None):
     Returns
     -------
     h_with_memory : WaveformModes
-       WaveformModes object corresponding to the strain with electric memory
+        WaveformModes object corresponding to the strain with electric memory
     psi4_with_memory : WaveformModes, optional
-       WaveformModes object corresponding to `psi4` with electric memory.  If 
-       `psi4` is `None`, then this is absent.
+        WaveformModes object corresponding to `psi4` with electric memory.  If
+        `psi4` is `None`, then this is absent.
 
     """
     h_memory_correction = J_E(h, integration_start_time=integration_start_time)
--- a/sxs/waveforms/memory.py
+++ b/sxs/waveforms/memory.py
@@ -237,10 +237,11 @@ def J_J(h):
     return J_𝒥
 
 
-def add_memory(h, integration_start_time=None):
-    """Add electric component of null memory to strain
+def add_memory(h, integration_start_time=None, psi4=None):
+    """Add electric component of null memory to strain and optionally psi4
 
-    This adds the contribution from the energy flux to the strain.
+    This adds the contribution from the energy flux to the strain, and
+    optionally adds minus the 2nd derivative of this contribution to psi4.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


sxs/waveforms/memory.py
```diff
@@ -249,13 +250,23 @@ def add_memory(h, integration_start_time=None):
...
+        WaveformModes object corresponding to psi4
 
     Returns
     -------
-    h_with_memory : WaveformModes
-        WaveformModes object corresponding to the strain with electric memory
+    (h_with_memory, psi4_with_memory) : tuple of WaveformModes
+       WaveformModes objects corresponding to the strain with electric memory
+       and psi4 with electric memory.  If psi4 is None, then psi4_with_memory
+       is also None.
```
moble, line 261: For backwards compatibility, we should keep the same return type (just the single waveform) when psi4 is None

```suggestion
    h_with_memory : WaveformModes
       WaveformModes object corresponding to the strain with electric memory
    psi4_with_memory : WaveformModes, optional
       WaveformModes object corresponding to `psi4` with electric memory.  If 
       `psi4` is `None`, then this is absent.
```

sxs/waveforms/memory.py
```diff
@@ -249,13 +250,23 @@ def add_memory(h, integration_start_time=None):
...
 
     """
-    h_with_memory = WaveformModes(MTS(h) + J_E(h, integration_start_time=integration_start_time))
+    h_memory_correction = J_E(h, integration_start_time=integration_start_time)
+    h_with_memory = WaveformModes(MTS(h) + h_memory_correction)
     h_with_memory.register_modification(add_memory, integration_start_time=integration_start_time)
-    return h_with_memory
+    if psi4 is None:
+        return (h_with_memory, None)
```
moble, line 268:
```suggestion
        return h_with_memory
```

moble (changes requested):

sxs/waveforms/memory.py
```diff
@@ -249,13 +250,25 @@ def add_memory(h, integration_start_time=None):
...
+       WaveformModes objects corresponding to the strain with electric memory
+       and psi4 with electric memory.  If psi4 is None, then psi4_with_memory
+       is also None.
 
     """
-    h_with_memory = WaveformModes(MTS(h) + J_E(h, integration_start_time=integration_start_time))
-    h_with_memory.register_modification(add_memory, integration_start_time=integration_start_time)
-    return h_with_memory
+    h_memory_correction = J_E(h, integration_start_time=integration_start_time)
+    h_with_memory = WaveformModes(MTS(h) + h_memory_correction)
```
keefemitman, line 265: Do we want to copy over the metadata as well? E.g.,
`h_with_memory.metadata = h.metadata`

moble: Note the difference between `metadata` and `_metadata`.  The latter contains all the info like the time array, `ell_min`, `ell_max`, data type, etc., and is copied over automatically already.  On the other hand `h.metadata` might not even be present; it contains information from the metadata.json/txt file, if it happened to be in the same directory when the waveform was loaded.  It's more for interactive convenience than correctness, so I think it should be dropped even if it's present in the original waveform.

keefemitman: Ah ok I see, thanks for the clarification

markscheel:
Made changes and squashed.

moble (approved):

moble:
Thanks Mark.  I'm just gonna make another little change to the versions listed in the JSON file, and then as soon as the changes hit conda-forge, you should be good to go.

PR #66, opened 2022-07-19
Add PN generator

dongzesun:
Add corrected PostNewtonion generator, only Taylor T1 is included for now.


moble:
As we had discussed previously, all of the notebooks and such should go in [the original `PostNewtonian` repo](https://github.com/moble/PostNewtonian).  Please create that and a new PR in this repo with just the resulting code files that will create a PN waveform.

PR #75, opened 2023-05-12
Fix normalization bug in align2d script

keefemitman:
sxs's norm is defined as the sqrt of the L^2 norm, but in the align2d code we want the actual L^2 norm.

### DIFF
--- a/sxs/waveforms/alignment.py
+++ b/sxs/waveforms/alignment.py
@@ -262,7 +262,7 @@ def align2d(wa, wb, t1, t2, n_brute_force_δt=None, n_brute_force_δϕ=5, includ
 
     normalization = trapezoid(
         CubicSpline(
-            wb.t, wb[:, wb.index(2, -2) : wb.index(ell_max + 1, -(ell_max + 1))].norm
+            wb.t, wb[:, wb.index(2, -2) : wb.index(ell_max + 1, -(ell_max + 1))].norm ** 2
         )(t_reference),
         t_reference,
     )


moble:
I think I agree with the actual code change, but not with your comment above.

First, the normalization should be the sum of squares (without a square-root) because [it divides something](https://github.com/sxs-collaboration/sxs/blob/201ac28b2cc17e21068223cc7441b7672665bcc9/sxs/waveforms/alignment.py#L128-L134) that is explicitly constructed as the sum of squares.  So I agree that they should be on the same footing, and your change to the code is correct.  But maybe I'm missing something because I believe that `WaveformModes.norm` is defined as the $L^2$ norm, *not* its square-root.  Maybe [the documentation of `WaveformModes.norm`](https://github.com/sxs-collaboration/sxs/blob/201ac28b2cc17e21068223cc7441b7672665bcc9/sxs/waveforms/waveform_modes.py#L244-L266) is poorly written?

I'll walk through my reasoning carefully, for myself as much as anything.  There are some overlapping meanings for the notation $L^2$, but generally they all include squaring, then summing/integrating, and then taking a square-root.  There is an annoying ambiguity with the term "norm" [in the context of quaternions](https://github.com/moble/quaternionic#attributes) (and other composition algebras), where it might be more specifically called the "Cayley norm", and there is no square root.  But AFAICT otherwise, a 2-"norm" always involve a square-root.  In particular, the underlying concept here is the space of square-integrable functions, for which the $L^2$ norm is universally defined as

$$ \\|f\\|_2 := \left( \int |f|^2 d\Omega \right)^{1/2} $$

We can expand $f$ in SWSHs, and then use orthonormality to find

$$
\begin{align}
  \\|f\\|_2 &= \left( \int \left| \sum\_{\ell,m} f\_{\ell,m} \\, {}\_sY\_{\ell,m} \right|^2 d\Omega \right)^{1/2} \\
   &= \left( \int \left( \sum\_{\ell,m} f\_{\ell,m} \\, {}\_sY\_{\ell,m} \right)\left( \sum\_{\ell',m'} \bar{f}\_{\ell',m'} \\, {}\_s\bar{Y}\_{\ell',m'} \right) d\Omega \right)^{1/2} \\
   &= \left( \sum\_{\ell,m,\ell',m'} f\_{\ell,m} \bar{f}\_{\ell',m'} \int {}\_sY\_{\ell,m} \\, {}\_s\bar{Y}\_{\ell',m'} d\Omega \right)^{1/2} \\
   &= \left( \sum\_{\ell,m,\ell',m'} f\_{\ell,m} \bar{f}\_{\ell',m'} \delta\_{\ell,\ell'} \delta\_{m,m'} \right)^{1/2} \\
   &= \left( \sum\_{\ell,m} f\_{\ell,m} \bar{f}\_{\ell,m} \right)^{1/2} \\
   &= \left( \sum\_{\ell,m} \left| f\_{\ell,m} \right|^2 \right)^{1/2}
\end{align}
$$

This last quantity is what `np.linalg.norm` returns, which is what [`WaveformModes.norm`](https://github.com/sxs-collaboration/sxs/blob/201ac28b2cc17e21068223cc7441b7672665bcc9/sxs/waveforms/waveform_modes.py#L266) calls.

Is this consistent with your understanding of the code and/or documentation?

keefemitman:
@moble Yeah my apologies for using sloppy wording. I think the confusion mainly arises because `sxs.WaveformModes.norm` is defined as $`||f||_{2}`$, i.e., the $`L^{2}`$ norm, as you've written out explicitly.

`scri.WaveformModes.norm`, however, is defined as $`(||f||_{2})^{2}`$, since the default of `take_sqrt` is `take_sqrt=False`.

It would be nice if the two were consistent.

moble:
Ah, okay.  I forgot about that `take_sqrt` argument.  That was an ugly choice, alright...

I wouldn't be opposed to making a breaking release of `scri` to change that, but I'm not sure if anyone uses that in existing code.  There are a few uses within `scri` that would have to be changed.  Thoughts?

For now, anyway, I'll just merge this.  Thanks.

keefemitman:
Yeah I agree that it's too big of a change (and not important enough of a change) to make to `scri`. Thanks for merging!

Issue #85, opened 2024-02-28
`sxs` is undefined for when calling `ellhat`


SSL32081:
I was trying to call the `ellhat` property of the `horizons` object, 
```
horizons = sxs.load("SXS:BBH:1593/Lev/Horizons")
horizons.ellhat
```
but I got an error:
```
NameError: name 'sxs' is not defined
```

PR #86, opened 2024-02-28
Attempt to fix `sxs` undefined.

SSL32081:
This pull request is trying to fix the issue #85.

----
📚 Documentation preview 📚: https://sxs--86.org.readthedocs.build/en/86/


moble:
Thank you!  And sorry about that.

I've actually found another few cases where that sort of thing has crept into the code, so I'll fix those soon, too.

Issue #48, opened 2021-12-20
Imported property 'bar' is not callable


hrueter:
In the `spherical` package 'bar' is a property, but it seems the implementation in `sxs` treats it as a function, which leads to the following error:
```
File "/home/hannes/.local/lib/python3.8/site-packages/sxs/waveforms/waveform_modes.py", line 288, in bar
    return spherical.modes.algebra.bar(self)
TypeError: 'property' object is not callable
```

moble:
Fixed by #87

PR #87, opened 2024-03-05
Fix `bar` is not a property

SSL32081:
This pull request attempts to address issue #48, where `bar` is not a function to be called. Use the original `conjugate` function define in `spherical.modes` instead.

This PR also fixes a small `frame` undefined issue.

----
📚 Documentation preview 📚: https://sxs--87.org.readthedocs.build/en/87/


moble:
Thanks!

PR #88, opened 2024-03-13
Update convert_simulation to use new catalog format

keefemitman:
Update convert_simulation to use new sxs catalog format.

----
📚 Documentation preview 📚: https://sxs--88.org.readthedocs.build/en/88/


sxs/utilities/lvcnr/conversion.py
```diff
@@ -145,9 +145,14 @@ def convert(self, sxs_data_path, out_path, truncation_time=None, resolution=None
...
-        with open(os.path.join(sxs_data_path, "metadata.json"), 'r') as f:
-            metadata = json.load(f)
+        # Load metadata from this simulation
+        try:
+            metadata = sxs.Metadata.from_file(os.path.join(sxs_data_path, "metadata.txt"))
+        except:
+            try:
+                metadata = sxs.Metadata.from_file(os.path.join(sxs_data_path, "metadata.json"))
+            except:
+                raise ValueError('Cannot load metadata.')
```
moble, line 155: This would prefer `metadata.txt`, but I think it makes more sense to prefer `metadata.json`, in which case the `from_file` function chooses automatically:

```suggestion
        try:
            metadata = sxs.Metadata.from_file(os.path.join(sxs_data_path, "metadata"))
        except:
            raise ValueError('Cannot load metadata.')
```

sxs/utilities/lvcnr/conversion.py
```diff
@@ -165,7 +170,7 @@ def convert(self, sxs_data_path, out_path, truncation_time=None, resolution=None
         log("Output filename is '{0}'".format(out_name))
 
         start_time, peak_time, version_hist = convert_modes(
-            sxs_data_path + "/rhOverM_Asymptotic_GeometricUnits_CoM.h5",
+            sxs_data_path + "/Strain_N2",
```
moble, line 173: In case — e.g., Geoffrey — uses this at some point, could you add logic like the above to prefer Strain_N2, but if it doesn't exist choose rhOverM?

sxs/utilities/lvcnr/waveforms.py
```diff
@@ -19,15 +20,22 @@ def convert_modes(sxs_format_waveform, metadata, out_filename,
...
-    mode_strings = ["Y_l{0[0]}_m{0[1]}.dat".format(ellm) for ellm in modes]
-    h = WaveformAmpPhase.read(sxs_format_waveform, extrap, mode_strings=mode_strings, start_time=truncation_time)
+    h = sxs.load(sxs_format_waveform)
+    if not truncation_time is None:
+        h = h[np.argmin(abs(h.t - truncation_time)) + 1:]
 
     start_time = h.t[0]
 
-    peak_time = h.t[h.peak_index()]
+    peak_time = h.t[np.argmax(h.norm)]
```
moble, line 29:
```suggestion
    peak_time = h.max_norm_time()
```

sxs/utilities/lvcnr/waveforms.py
```diff
@@ -47,4 +55,4 @@ def convert_modes(sxs_format_waveform, metadata, out_filename,
             out_group_phase = out_file.create_group('phase_l{0[0]}_m{0[1]}'.format(mode))
             phase_out.write(out_group_phase)
 
-    return start_time, peak_time, h.version_hist
+    return start_time, peak_time, None
```
moble, line 58:
```suggestion
    return start_time, peak_time, getattr(h, "version_hist", None)
```

sxs/utilities/lvcnr/waveforms.py
```diff
@@ -19,15 +20,22 @@ def convert_modes(sxs_format_waveform, metadata, out_filename,
...
-    peak_time = h.t[h.peak_index()]
+    peak_time = h.t[np.argmax(h.norm)]
 
     t = h.t - peak_time
 
+    h_amp_phase = np.empty((h.data.shape[0], 2*h.data.shape[1]))
+    for mode in modes:
+        L, M = mode
+        h_amp_phase[:,2*sf.LM_index(L, M, h.ell_min)] = np.abs(h.data[:,sf.LM_index(L, M, h.ell_min)])
+        h_amp_phase[:,2*sf.LM_index(L, M, h.ell_min) + 1] = np.unwrap(np.angle(h.data[:,sf.LM_index(L, M, h.ell_min)]))
```
moble, line 37:
```suggestion
h_amp_phase = np.empty((h.data.shape[0], 2*h.data.shape[1]))
h_amp_phase[:,::2] = h.abs
h_amp_phase[:,1::2] = h.arg_unwrapped
```

sxs/utilities/lvcnr/waveforms.py
```diff
@@ -2,7 +2,8 @@
 
 import numpy as np
 import h5py
-
+import sxs
+import spherical_functions as sf
```
moble, line 6: Instead of `sf.LM_index(L, M, h.ell_min)`, you should just use `h.index(L, M)`, so you shouldn't need such an import at all.  (But also, `spherical_functions` is the C-style dependency; we need to stick to `spherical` for `sxs.)

```suggestion
```

sxs/utilities/lvcnr/waveforms.py
```diff
@@ -2,7 +2,8 @@
 
 import numpy as np
 import h5py
-
+import sxs
```
moble, line 5: Doesn't this cause a circular import issue?  I think you probably want to use `from ... import load` (or so) inside the `convert_modes` function.

moble (changes requested):

sxs/utilities/lvcnr/conversion.py
keefemitman: I'll do this and also just make it an optional argument to provide a waveform name

sxs/utilities/lvcnr/conversion.py
```diff
@@ -141,64 +157,105 @@ def convert(self, sxs_data_path, out_path, truncation_time=None, resolution=None
...
-        out_name = out_path + "/" + sxs_id.replace(':', '_') + "_Res" + str(resolution) + ".h5"
+        if not out_waveform_name is None:
+            out_name = out_path + "/" + out_waveform_name
+        else:
+            out_name = out_path + "/" + sxs_id.replace(":", "_") + "_Res" + str(resolution) + ".h5"
         log("Output filename is '{0}'".format(out_name))
 
+        if waveform_name is None:
+            if os.path.exists(os.path.join(sxs_data_path, "Strain_N2.h5")):
+                waveform_path = sxs_data_path + "/Strain_N2"
```
moble, line 193:
```suggestion
                waveform_path = sxs_data_path + "/Strain_N2.h5"
```

keefemitman: No `sxs.load` actually seems to fail if you include the .h5 suffix...

moble: Hmm.  That sounds like a bug.  Any more detail?

keefemitman: It looks like RPDMB just can't find the .json file when the .h5 suffix is included; running `sxs.load('Strain_N2.h5')` for anything in the catalog fails, but `sxs.load('Strain_N2')` works fine. I can post the exact error message if you want.

moble: Is this just when you're working with git annex?  If so, I believe I have fixed that issue in my big PR, but for now we could just leave this as a hack.  I would still call it undefined behavior, so it could break in the future, but we could just cross that bridge when we come to it.  Or maybe after my PR merges, we could check again and fix it then.

moble: Ah, it'll probably be easier for me to rebase onto this, so I might as well just merge it now.  A functional but missing `.h5` probably won't be the smelliest part of this LVCNR stuff anyway...

PR #90, opened 2024-07-15
Add 4d SO3 optimizer

keefemitman:
Extend `align2d` to optimize over the full SO(3) transformation (as well as the time translation).

----
📚 Documentation preview 📚: https://sxs--90.org.readthedocs.build/en/90/


keefemitman:
After talking with @duetosymmetry I'll change the optimization space to be `log(q)`

sxs/waveforms/alignment.py
```diff
@@ -1,12 +1,13 @@
 import numpy as np
 
+import quaternion
```
duetosymmetry, line 3: Everything else in this package is using the package quaternionic instead of quaternion...

sxs/waveforms/alignment.py
```diff
@@ -310,3 +299,227 @@ def align2d(wa, wb, t1, t2, n_brute_force_δt=None, n_brute_force_δϕ=5, includ
...
+    )
+
+    wa_prime = wa_prime.rotate(δSO3.components)
+
+    # Take the sqrt because least_squares squares the inputs...
+    diff = trapezoid(
+        np.sum(abs(wa_prime.data - modes_B) ** 2, axis=1),
+        t_reference,
+    )
+    return np.sqrt(diff / normalization)
```
moble, line 329: If you're passing this to `least_squares`, you should just return the residuals, rather than summing or integrating them — and certainly not the sqrt.  Normalization is fine, and you'll probably want to reinterpret the complex difference with `.view(float)`.

sxs/waveforms/alignment.py
```diff
@@ -310,3 +299,227 @@ def align2d(wa, wb, t1, t2, n_brute_force_δt=None, n_brute_force_δϕ=5, includ
...
+        raise ValueError(f"(t1,t2)=({t1}, {t2}) not contained in wb.t, which spans ({wb.t[0]}, {wb.t[-1]})")
+
+    if max_δt is None:
+        max_δt = np.inf
+
+    # Figure out time offsets to try
+    δt_lower = min(max_δt, max(t1 - t2, wa.t[0] - t1))
+    δt_upper = min(max_δt, min(t2 - t1, wa.t[-1] - t2))
+
+    # We'll start by brute forcing, sampling time offsets evenly at as many
```
moble, line 438: I think we need to rethink the approach of this function, starting here.  The name of the game is to get good initial guesses, but the curse of dimensionality means that brute-force methods are that much worse.   So it would probably be better to optimize hierarchically.  I think the way to go would be to

  1. Run `align1d` to get the initial guess for the time offset `δt`.  The cost function here is invariant under rotations, so the crazy precession that this function should handle shouldn't be an problem.  And I would guess that `δt` will be a great guess for the final result from this function.
  2. Apply that `δt`, then measure the *unit* angular velocity vector for each — within (t1,t2) — and use `quaternionic.align` (only available in v1.0.11) to get the rotor `R` that will best align their angular velocities.  For precessing systems, that could be enough to fully align them, but either way, there's no harm in checking the rotation about that axis.
  3. Using `δt` and `R`, brute-force optimize `R*exp(theta*z/2)` for `4*ell_max+2` values of `theta` in [0,2*pi).
  4. Use those values as the starting point for the active 4D optimization, which should be pretty easy and quick.

sxs/waveforms/alignment.py
```diff
@@ -310,3 +297,226 @@ def align2d(wa, wb, t1, t2, n_brute_force_δt=None, n_brute_force_δϕ=5, includ
...
 
     return optimums[idx].cost, wa_primes[idx], optimums[idx]
+
+
+def _cost4d(δt_δSO3, args):
+    from .. import WaveformModes
+
+    modes_A, modes_B, t_reference, normalization = args
+    δt = δt_δSO3[0]
+    δSO3 = np.exp(np.quaternion(*δt_δSO3[1:]))
```
moble, line 307:
```suggestion
def _cost4d(δt_δso3, args):
    from .. import WaveformModes

    modes_A, modes_B, t_reference, normalization = args
    δt = δt_δso3[0]
    δSO3 = np.exp(quaternionic.array.from_vector_part(δt_δso3[1:]))
```

sxs/waveforms/alignment.py
```diff
@@ -310,3 +297,226 @@ def align2d(wa, wb, t1, t2, n_brute_force_δt=None, n_brute_force_δϕ=5, includ
     idx = np.argmin(abs(np.array([optimum.cost for optimum in optimums])))
 
     return optimums[idx].cost, wa_primes[idx], optimums[idx]
+
+
+def _cost4d(δt_δSO3, args):
+    from .. import WaveformModes
+
```
moble, line 304: np.quaternion only exists if you import quaternion.  We really have to use quaternionic.  I'm not sure I'll even be able to maintain quaternion now that numpy 2.0 has broken everything.

moble (changes requested):

sxs/waveforms/alignment.py
keefemitman: I was just repeating the style of https://github.com/sxs-collaboration/sxs/blob/0b5182768956dced348173d7fad7951a6e751e80/sxs/waveforms/alignment.py#L109 and https://github.com/sxs-collaboration/sxs/blob/0b5182768956dced348173d7fad7951a6e751e80/sxs/waveforms/alignment.py#L123. Do you want to change those as well?

PR #99, opened 2024-08-17
spectre_cce_v1.py: fix missing ell_min

nilsvu:
I hit this bug when loading data in the spectre CCE format without the ell=0,1 modes.

----
📚 Documentation preview 📚: https://sxs--99.org.readthedocs.build/en/99/


moble:
Failures were due to connectivity issues; everything looks good.  Thanks!

PR #119, opened 2024-12-23
Allow Metadata to be updated with kwargs

prayush:
The current `Metadata` class disallows `update(...)` with keyword arguments, which its base class `OrderedDict` allows. One has to:

```
import sxs
x = sxs.Metadata()
x.update(dict(reference-time=100))
```
instead of 
```
x.update(reference-time=100)
```

This PR will enable the latter syntax, aligning the `Metadata` class closer to its base class.

----
📚 Documentation preview 📚: https://sxs--119.org.readthedocs.build/en/119/


duetosymmetry:
This seems like a nice idea — I think we should follow the pattern of the builtin container. The documentation for `dict.update` says:
```
Help on built-in function update:

update(...) method of builtins.dict instance
    D.update([E, ]**F) -> None.  Update D from dict/iterable E and F.
    If E is present and has a .keys() method, then does:  for k in E: D[k] = E[k]
    If E is present and lacks a .keys() method, then does:  for k, v in E: D[k] = v
    In either case, this is followed by: for k in F:  D[k] = F[k]
```
this means that the kwargs assignment should be done afterwards. And, I think we don't need to check isinstance — we can just do what the documentation suggests, `try`ing with `keys()`.

duetosymmetry:
@moble can you figure out what's wrong with the deps in the tests? It seems to work on some platforms but not others. It doesn't have anything to do with Prayush's code

moble:
Thanks guys!  Sorry for the delay.

PR #126, opened 2025-02-16
Correct factor of 2 in 4d optimization rotation

keefemitman:
In the original 4d optimization PR this line had factors of $\pi$ instead of $\pi/2$; I believe the former was correct since the angles are free to range from 0 to $2\pi$ (note the factor of "/ 2" in line 482)

----
📚 Documentation preview 📚: https://sxs--126.org.readthedocs.build/en/126/

### DIFF
--- a/sxs/waveforms/alignment.py
+++ b/sxs/waveforms/alignment.py
@@ -525,7 +525,7 @@ def align4d(
     optimum = least_squares(
         cost_wrapper,
         δt_δso3,
-        bounds=[(δt_lower, -np.pi/2, -np.pi/2, -np.pi/2), (δt_upper, np.pi/2, np.pi/2, np.pi/2)],
+        bounds=[(δt_lower, -np.pi, -np.pi, -np.pi), (δt_upper, np.pi, np.pi, np.pi)],
         max_nfev=50000,
     )
     δt = optimum.x[0]


moble:
No, the *rotation* angles should span $2\pi$, but the angles you're talking about should only span $\pi$ — as long as we're only working with fields with integer spin.  They are exponentiated into a quaternion $R$, which acts on a vector $\mathbf{v}$ as $R \mathbf{v} R^{-1}$.  The presence of two 𝑅s effectively doubles the angles.  To rotate a vector through an angle $\alpha$ about a unit vector $\mathbf{u}$, we need the quaternion $R = \exp[\alpha \mathbf{u} / 2]$, and it's that factor of $1/2$ you're seeing here.

Another way to see this is to note that $R \to -R$ has no effect on the *rotation* because of those two 𝑅s.<sup>(*)</sup> 
You can also check (e.g., in Eq. (33) of https://arxiv.org/pdf/1604.08140) that

$$\mathfrak{D}^{(\ell)}\_{m',m}(R) = \mathfrak{D}^{(\ell)}\_{m',m}(-R)$$

and thus

$${}\_sY_{\ell,m}(R) = {}\_sY_{\ell,m}(-R)$$

(for integer indices).  In general, for any number $\beta$ and unit vector $\mathbf{u}$, we have

$$ \exp\left[ \beta \mathbf{u} + \pi \mathbf{u} \right] = \exp\left[ \beta \mathbf{u}\right] \exp\left[ \pi \mathbf{u} \right] = -\exp\left[ \beta \mathbf{u} \right], $$

which has no effect on the rotation or function values (for integer indices).  Thus, we see that the logarithm's angles only need to span $\pi$, so ranging from $-\pi/2$ to $\pi/2$ will be sufficient (for integer indices).

---

<sup>(*)</sup>  This is just the statement that the unit quaternions $\mathrm{Spin}(3)$ form a double cover of the rotation group $\mathrm{SO}(3)$.  I probably should have insisted on variable names with `Spin3` instead of `SO3`.

moble:
> the code is failing because the initial guess is outside of the bounds

Ah, of course.  Subtle!

Closing in favor of #128.

Issue #125, opened 2025-02-15
Result of align2d doesn't have spin_weight


duetosymmetry:
https://github.com/sxs-collaboration/sxs/blob/8dfeebcade9568b88648e569df6ac19f5b96a60b/sxs/waveforms/alignment.py#L291-L302

Just add `spin_weight=spin_weight`

moble:
Closed by #129

PR #129, opened 2025-02-18
Correct align4d rotation bounds; missing spin_weight in align2d (#125)

duetosymmetry:
Bounds for the algebra element δso3 should be in the ball of radius pi, not pi/2.

Result of 2d alignment was coming back with no value for spin_weight.

----
📚 Documentation preview 📚: https://sxs--129.org.readthedocs.build/en/129/


moble:
The spin weight fix looks correct.  But the δso3 bounds should only be $\pm \pi/2$; see [this comment](https://github.com/sxs-collaboration/sxs/pull/126#issuecomment-2661528695).  Keefe did find a problem where the initial guess was outside the bounds, but I *believe* I fixed that in #128.  Did you find a problem using v2024.0.32?

duetosymmetry:
I hit the bug before the discussion you linked to, and hadn't seen the discussion. I had made the change locally (so I no longer hit the bug) and only committed+pushed today. Will try with v2024.0.32 sometime tomorrow, if I get a chance.

moble:
I want that spin weight fix, but if you really think there's a need for a span of 2pi, open another PR.

moble:
and thanks!

duetosymmetry:
No prob, sorry I didn't get a chance to test.

PR #130, opened 2025-02-20
Indentation error in returning metadata meant no return

duetosymmetry:
Indentation error in this return statement meant there was no return if taking any branches before the last one

----
📚 Documentation preview 📚: https://sxs--130.org.readthedocs.build/en/130/


moble:
Ah, sorry about that!  Thanks for the fix.

PR #138, opened 2025-03-03
Add simulation/waveform alignment functions to reproduce catalog analysis results

keefemitman:
@moble this is suppose to serve as the base for updating `sxs` to include the code necessary to reproduce the analysis that we're presenting in the catalog paper. I figured it'd be good for us to discuss things and merge this (or maybe some modified/edited version of it) first before I dig too deep into adding the more specific error functions, e.g., mismatches with respect to ASD curves or individual mode errors. Let me know your thoughts and if you'd like me to change anything. Otherwise we can also discuss this during the vacuum call.

### DIFF
--- a/sxs/waveforms/alignment.py
+++ b/sxs/waveforms/alignment.py
@@ -7,6 +7,8 @@ from scipy.integrate import trapezoid
 import multiprocessing as mp
 from functools import partial
 
+from . import norms
+
 
 def align1d(wa, wb, t1, t2, n_brute_force=None):
     """Align waveforms by shifting in time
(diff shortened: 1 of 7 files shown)


keefemitman:
Ok not sure what's going on with `align2d` but I'll look into this

sxs/waveforms/norms.py
```diff
@@ -0,0 +1,226 @@
...
+        Interpolated to wa over (t1, t2).
+    t1 : float
+        Beginning of L2 norm integral.
+    t2 : float
+        End of L2 norm integral.
+    modes : list, optional
+        Modes (\ell, m) to include in numerator of L2 norm calculation.
+        Default is all modes.
+    modes_for_norm : list, optional
+        Modes (\ell, m) to include in denominator of L2 norm calculation.
```
moble, line 78: The backslashes will just confuse python

sxs/waveforms/alignment.py
```diff
@@ -549,3 +551,284 @@ def align4d(
     wa_prime = wa_prime.rotate(δSpin3)
 
     return optimum.cost, wa_prime, optimum
+
+def map_waveform_to_canonical_frame(wa, t_ref):
+    """Map waveform to canonical frame at t_ref; this amounts to
```
moble, line 556: Docstrings are supposed to have a single-line summary, followed by two newlines, then more description if needed.  This format will break some help tools.

sxs/waveforms/alignment.py
```diff
@@ -549,3 +551,284 @@ def align4d(
     wa_prime = wa_prime.rotate(δSpin3)
 
     return optimum.cost, wa_prime, optimum
+
+def map_waveform_to_canonical_frame(wa, t_ref):
```
moble, line 555: I wonder if this could share could with [the stuff in `to_lvc_conventions`](https://github.com/sxs-collaboration/sxs/blob/7069e94f943e62ffe6fed4f54de708e24c29d6d0/sxs/waveforms/format_handlers/lvc.py#L105-L159).  Or at least, they should reference each other in a "See Also" section.

moble (changes requested):
Didn't fully review, but you asked for a quick look to see if it makes sense.  Makes sense.

sxs/waveforms/norms.py
```diff
@@ -0,0 +1,249 @@
...
+        ASD_values = np.ones_like(wa.t)
+
+    if type(wa) == _ModesTimeSeries:
+        overlap = trapezoid(
+            np.sqrt(4 * np.pi) * wa.multiply(wb.bar, truncator=lambda tup: 0).ndarray[:, 0] / ASD_values**2,
+            wa.t,
+        ).real
+    else:
+        trapezoid(
+            (wa * wb.bar).ndarray / ASD_values**2,
```
moble, line 174: If this is just a `TimeSeries`, isn't `.bar` undefined?

sxs/waveforms/norms.py
```diff
@@ -0,0 +1,249 @@
...
+    if ASD_values is None:
+        ASD_values = np.ones_like(wa.t)
+
+    if type(wa) == _ModesTimeSeries:
+        overlap = trapezoid(
+            np.sqrt(4 * np.pi) * wa.multiply(wb.bar, truncator=lambda tup: 0).ndarray[:, 0] / ASD_values**2,
+            wa.t,
+        ).real
+    else:
+        trapezoid(
```
moble, line 173:
```suggestion
        overlap = trapezoid(
```

sxs/waveforms/norms.py
```diff
@@ -0,0 +1,249 @@
...
+            (wa * wb.bar).ndarray / ASD_values**2,
+            wa.t,
+        ).real
+
```
moble, line 177: Then we'd need an `else` here to raise an error.

sxs/waveforms/norms.py
```diff
@@ -0,0 +1,249 @@
...
+    Returns
+    -------
+    wa_interp : WaveformModes
+    wa_interp : WaveformModes
+    """
+    check_time_constraint(wa, wb, t1, t2)
+
+    padding_time = (t2 - t1) * padding_time_factor
+    t1_padded = max(max(wa.t[0], wb.t[0]), t1 - padding_time)
+    t2_padded = min(min(wa.t[-1], wb.t[-1]), t2 + padding_time)
```
moble, line 48:
```suggestion
    t1_padded = max(wa.t[0],  wb.t[0],  t1 - padding_time)
    t2_padded = min(wa.t[-1], wb.t[-1], t2 + padding_time)
```

sxs/waveforms/waveform_modes.py
```diff
@@ -1275,6 +1280,69 @@ def preprocess(self, t1=0, t2=500, t3=None, t4=None, dt=None, **kwargs):
...
+        h_tilde : TimeSeries
+            Dimensionful Fourier transform.
+        """
+        if theta_phi is not None:
+            h_dimensionful = self.evaluate(*theta_phi)
+            h_dimensionful *= total_mass * h_factor
+            h_dimensionful.t *= total_mass * t_factor
+        else:
+            h_dimensionful = WaveformModes(
+                input_array=self.data * total_mass * h_factor,
```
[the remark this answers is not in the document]
keefemitman: This is just making the strain dimensionful before FFTing. Maybe I'm misunderstanding you?

sxs/waveforms/waveform_modes.py
```diff
@@ -1275,6 +1280,69 @@ def preprocess(self, t1=0, t2=500, t3=None, t4=None, dt=None, **kwargs):
...
+                time=self.t * total_mass * t_factor,
+                time_axis=self.time_axis,
+                modes_axis=self.modes_axis,
+                ell_min=self.ell_min,
+                ell_max=self.ell_max,
+                spin_weight=self.spin_weight,
+            )
+
+        N = h_dimensionful.t.size
+        max_dt = max(np.diff(h_dimensionful.t))
```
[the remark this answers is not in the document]
keefemitman: I was assuming that the user would have already done this post-processing, e.g., running `h.preprocess`. Do you think we should make it an option to run the `preprocess` function in `fourier_transform`?

sxs/waveforms/waveform_modes.py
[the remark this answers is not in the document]
moble: Yeah, it's that part about "up to the luminosity distance" that worries me.  I can see an argument for not worrying about units / scaling at all because we'll be normalizing things anyway.  And I can see an argument for scaling by some $M$ because the native total mass isn't quite 1, for example.  But I don't see an argument for leaving the strain dimensionful.

moble (changes requested):

sxs/waveforms/waveform_modes.py
```diff
@@ -1275,6 +1276,83 @@ def preprocess(self, t1=0, t2=500, t3=None, t4=None, dt=None, **kwargs):
...
+                fft.fftshift(
+                    fft.fft(h_unitful.data, axis=self.time_axis), axes=self.time_axis
+                ),
+                fft.fftshift(fft.fftfreq(N, max_dt)),
+            )
+        else:
+            h_tilde = WaveformModes(
+                input_array=fft.fftshift(
+                    fft.fft(h_unitful.data, axis=self.time_axis), axes=self.time_axis
+                ),
```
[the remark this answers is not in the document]
keefemitman: Ah I see what you mean now! Yes this is what it should be

sxs/simulations/analyze.py
```diff
@@ -0,0 +1,265 @@
...
+
+    wa, wb = create_unified_waveforms(wa, wb, t1, t2, padding_time_factor=0)
+
+    errors["mismatch"] = compute_mismatch(wa, wb, t1, t2)
+
+    errors["residual L2 norm"] = compute_L2_norm(wa, wb, t1, t2)
+
+    f1 = (np.gradient(-np.unwrap(np.angle(wa.data[:, wa.index(2, 2)])), np.diff(wa.t * m_sun_in_seconds)[0]))[
+        np.argmin(abs(wa.t - t1))
+    ] / (2 * np.pi)
```
moble, line 57: Shouldn't this come after `preprocess`, since the waveforms aren't likely to have uniform time steps yet?  Or should the second argument to `gradient` just be the whole `wa.t * m_sun_in_seconds`?

sxs/waveforms/alignment.py
```diff
@@ -549,3 +544,332 @@ def align4d(
...
+    corresponding optimization is performed over [t1, t2].
+
+    Parameters
+    ----------
+    sima : Simulation
+    simb : Simulation
+        Simulations to be aligned
+    t1 : float
+        Beginning of integration interval.
+        Default is the relaxation time of simb.
```
moble, line 788:
```suggestion
        Default is the reference time of simb.
```

sxs/waveforms/alignment.py
```diff
@@ -549,3 +544,332 @@ def align4d(
...
+        L² norm of normalized residual between wa_prime and wb over [t1, t2].
+    t1 : float
+    t2 : float
+    """
+    wa = sima.h.copy()
+    wb = simb.h.copy()
+
+    if t1 is None:
+        # Default to relaxation time
+        t1 = simb.metadata.relaxation_time
```
moble, line 841:
```suggestion
        # Default to reference time
        t1 = simb.metadata.reference_time
```

sxs/simulations/analyze.py
moble: The more I look at this, the more I'm skeptical of it.  If there's noise or precession, this can go wonky.  A more robust approach is to just use the angular_velocity.  Something like this (with units adjusted appropriately; see other comments) would make me happier:
```suggestion
    i1 = wa.index_closest_to(t1)
    i0, i2 = max(0, i1-5), min(i1+6, wa.n_times-1)
    Ω1 = np.linalg.norm(h[i0:i2].angular_velocity, axis=1)[5]
    f1 = 2*Ω1 / (2*np.pi)
```

sxs/simulations/analyze.py
```diff
@@ -0,0 +1,265 @@
...
+    errors["t1"] = t1
+    errors["t2"] = t2
+
+    wa, wb = create_unified_waveforms(wa, wb, t1, t2, padding_time_factor=0)
+
+    errors["mismatch"] = compute_mismatch(wa, wb, t1, t2)
+
+    errors["residual L2 norm"] = compute_L2_norm(wa, wb, t1, t2)
+
+    f1 = (np.gradient(-np.unwrap(np.angle(wa.data[:, wa.index(2, 2)])), np.diff(wa.t * m_sun_in_seconds)[0]))[
```
moble: Is this the intended scaling?  Will the total mass be correctly included elsewhere?

sxs/simulations/analyze.py
```diff
@@ -0,0 +1,265 @@
...
+    ] / (2 * np.pi)
+
+    wa = wa.preprocess(t1, t1 + 0.01 * (t2 - t1), t2 - (t2 - t1) * 0.01, t2)
+    wb = wb.preprocess(t1, t1 + 0.01 * (t2 - t1), t2 - (t2 - t1) * 0.01, t2)
+
+    wa_tilde = wa.fourier_transform()
+    wb_tilde = wb.fourier_transform()
+    if ASDs is not None:
+        if total_masses is None:
+            total_masses = [1.0]
```
moble, line 66: I don't see why a total mass of 1 would be useful.  If nothing else, it seems to conflict with the scaling of `f1`, because it would really be a system of total mass 1 Msun, which we don't expect to exist.  Should this just raise an error?

sxs/simulations/analyze.py
```diff
@@ -0,0 +1,265 @@
...
+        for ASD in ASDs:
+            for total_mass in total_masses:
+                wa_tilde_total_mass = wa_tilde.copy()
+                wb_tilde_total_mass = wb_tilde.copy()
+                wa_tilde_total_mass.t /= total_mass
+                wb_tilde_total_mass.t /= total_mass
```
moble, line 72: I've lost track of the units by this point.  Could you check to make sure that they're correct, and add some explanatory comments?

moble (changes requested):
Also, please fix the test failures; they appear to be real.

sxs/simulations/analyze.py
keefemitman: I had originally done this so that the user could provide ASDs without a mass, but I agree this doesn't make much sense, so I'll make it raise an error.

sxs/simulations/analyze.py
keefemitman: I've added in the scaling which was indeed missing

sxs/simulations/analyze.py
keefemitman: Comments added

keefemitman:
Ok fixed a few more things @moble, ready for another round

moble:
Rebased and added a couple commits.  I  was worried about inconsistent times and LM modes, so I just went through making sure that everything would be made consistent.  Also changed function names
```
compute_L2_norm -> L2_difference
compute_inner_product -> inner_product
```
I'm happy with this now, so take a look and I'll merge when you're happy.

keefemitman:
@moble looks like `sxs.load(sxs_id, extrapolation="N3").h` (or `Outer`) no longer works? Did the convention change?

moble:
Sorry.  Try it now.

keefemitman:
@moble ok everything looks good to me now! 👍

PR #140, opened 2025-03-22
Attach session to function sxs.utilities.inspire.query, for reuse

duetosymmetry:
If the user performs many INSPIRE queries (e.g. in a loop), we should reuse the same session.  If the TCP session is still open, this saves time.  And in principle, if INSPIRE uses session-related cookies, we should persist those when possible (though I don't know if they do)

----
📚 Documentation preview 📚: https://sxs--140.org.readthedocs.build/en/140/

### DIFF
--- a/sxs/utilities/inspire.py
+++ b/sxs/utilities/inspire.py
@@ -44,7 +44,7 @@ def inspire2doi(inspire_bibtex_key, raise_exceptions=False):
 
 
 def query(
-    query,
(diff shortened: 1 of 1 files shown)


duetosymmetry:
Readthedocs test failure is unrelated (something to do with griffe having trouble, it throws `KeyError: 'Metadata'`)

moble:
Thanks!

PR #144, opened 2025-04-14
Fix ASD bug and allow for different total masses

keefemitman:
A minor change to the way ASDs are loaded by `sxs.simulations.analyze`. Fixes a bug that so the ASD-weighted mismatches are actually computed and enables ASDs to be evaluated against with different total masses, e.g., for LISA.

----
📚 Documentation preview 📚: https://sxs--144.org.readthedocs.build/en/144/


sxs/simulations/analyze.py
```diff
@@ -202,6 +204,26 @@ def analyze_simulation(
 
         errors[f"(-h.ddot, psi4)"] = L2_norm
```
moble, line 205:
```suggestion
        errors[f"(-h.ddot, psi4)"] = L2_norm
```

sxs/simulations/analyze.py
```diff
@@ -154,17 +156,17 @@ def analyze_simulation(
     sim = load(sim_name)
 
     # Lev analysis
-    if analyze_levs:
+    if analyze_levs and not len(sim.lev_numbers) < 2:
         for i, lev in enumerate(sim.lev_numbers[1:][::-1]):
```
moble, line 160: You still have some `lev - 1`s hanging around in this block.  I suggest just iterating over pairs:
```suggestion
        for i, (low_lev, high_lev) in enumerate(zip(sim.lev_numbers[:-1], sim.lev_numbers[1:])):
```
In this case, your `i == 0` condition will need to change to `i == len(sim.lev_numbers)-2`.

sxs/simulations/analyze.py
```diff
@@ -65,11 +68,9 @@ def compute_error_summary(wa, wb, t1, t2, modes=None, ASDs=None, total_masses=No
 
     wa_tilde = wa.fourier_transform()
     wb_tilde = wb.fourier_transform()
-    if ASDs is not None:
-        if total_masses is None:
-            raise ValueError("Need to specify total masses if ASDs are provided.")
-        for ASD in ASDs:
-            for total_mass in total_masses:
+    if ASDs_and_total_masses is not None:
+        for ASD, ASD_and_total_masses in ASDs_and_total_masses.items():
```
moble, line 72: Maybe change the names to be a little more descriptive, including unpacking:
```suggestion
        for ASD_name, (ASD_data, total_masses) in ASDs_and_total_masses.items():
```

PR #150, opened 2025-05-07
Updates for the 2025 catalog

moble:
- [x] Get readthedocs working again
- [x] Finish the `cite` function
- [x] Update docs
    - [x] Focus on the dataframe instead of simulations/metadata
	- [x] Clarify how to select Lev or extrapolation order
- [x] Remove *mention* of the `sxs[ecosystem]` option; leave it as an option for backwards compatibility
- [x] Run `hatch env prune && hatch run test --run_many_downloads --run_slow_tests` locally before final merge


duetosymmetry:
In sxs/simulations/simulation.py lines 48-49, we should also update from
```
     The returned object will be either a `Simulation_v1` or
     `Simulation_v2` object, depending on the version number.
```
to
```
    The returned object will be either a `Simulation_v1`, `Simulation_v2`,
    or `Simulation_v3` object, depending on the version number.
```

PR #151, opened 2025-05-08
SimulationBase.load_metadata should add computed md fields

duetosymmetry:
less-than-maximal Lev and older version simulations have `metadata=None` passed to
`Simulation_vN.__init__()`, which makes `__init__` call `load_metadata()`. But this
results in different behavior for the metadata loaded for less-than-maximal Lev
and older versions, since their metadata object is not constructed the same way
as in `load("simulations")[simulation_id]`.

Try to make it more uniform by using the `sxscatalog` functions
`_backwards_compatibility()` and `.add_standard_parameters()`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ classifiers = [
   "Topic :: Scientific/Engineering :: Astronomy"
 ]
 dependencies = [
-  "sxscatalog >=3.0.13",
+  "sxscatalog >=3.0.14",
   "numpy >=1.25",
   "scipy >=1.13",
   "numba >=0.55; implementation_name == 'cpython'",
--- a/sxs/simulations/simulation.py
+++ b/sxs/simulations/simulation.py
@@ -520,7 +520,9 @@ class SimulationBase:
         metadata_location = self.files.get(metadata_path)["link"]
         sxs_id_path = Path(self.sxs_id)
         metadata_truepath = Path(sxs_path_to_system_path(sxs_id_path / metadata_path))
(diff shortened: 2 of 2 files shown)


duetosymmetry:
@moble Two comments:
1. I don't love `from sxscatalog.metadata.metadata import _backwards_compatibility` -- probably would be cleaner to attach `_backwards_compatibility` to the object. I didn't do that because that requires changing two package. Will wait for your opinion.
2. This feels like duct tape instead of enforcing a uniform treatment of Metadata...

duetosymmetry:
Actually, a third comment. The reason I came across this was because `SimulationBase.__repr()__` assumes that `number_of_orbits` is a valid metadata field.

moble:
>     1. I don't love `from sxscatalog.metadata.metadata import _backwards_compatibility` -- probably would be cleaner to attach `_backwards_compatibility` to the object. I didn't do that because that requires changing two package. Will wait for your opinion.

I've included backwards compatibility in `add_standard_parameters` as of sxscatalog v3.0.14, so if you just update that dependency in pyproject.toml, you can remove the explicit call to `_backwards_compatibility`.


>     2. This feels like duct tape instead of enforcing a uniform treatment of Metadata...

Very much so; all of Metadata is a big ball of duct tape to deal with the fact that there has never previously been any standard at all for what metadata is included or how it's provided.  Because we need to be able to load old metadata files, I don't see this changing.

duetosymmetry:
Ok, I think this update ought to work then!

moble:
Thanks!

PR #165, opened 2025-06-11
rpdmb.save: L2norm_fractional_tolerance

### LINKED ISSUE #162: bugs with sxs.rpdmb.save's L2norm_fractional_tolerance parameter
I found a few problems:

1. The docs say you can supply `L2norm_fractional_tolerance=0.0`, but when I do that, it gives division-by-zero errors.
2. The code has statements `if L2norm_fractional_tolerance is None:`, which implies that None is a valid value for that parameter (and the docs say it is an optional parameter).  But when I try passing None, I get strange errors like `ValueError: Chunk shape must not be greater than data shape in any dimension. (23107, 1) is not compatible with (1, 3)`.
3. There appears to be no way of turning off `w.truncate()` so as to produce lossless compression with RPDMB, but with all the other steps in RPDMB unchanged. Normally one doesn't want lossless compression for production because the files are large, but I want it as a test.

I think the fundamental problem is that `L2norm_fractional_tolerance` is used in the code for multiple things:
1. As a tolerance parameter passed into `w.truncate()`
2. As a flag for bypassing the transformation to the corotating frame, if L2norm_fractional_tolerance is None (the docs don't mention this)
3. As a flag for bypassing the shuffle_widths , if  L2norm_fractional_tolerance is None (the docs say shuffle_widths are bypassed if L2norm_fractional_tolerance is 0.0, not None)
4. As a tolerance parameter passed into `w.to_corotating_frame()` (the docs don't mention this)
I think all of these should be controlled by different optional parameters, with defaults set to do what is done now.

All of this is with Python 3.12.2, `sxs.__version__='2025.0.16'`

markscheel:
Fixes #162

1) L2norm_fractional_tolerance is now used for w.truncate and nothing
   else.  Using None or values <=0 for this parameter means no
   truncation.
2) There is a new parameter corotating_frame_tolerance that
   is the tolerance passed into w.to_corotating_frame.
   It defaults to L2norm_fractional_tolerance for backward
   compatibility.
3) The code paths that formerly were taken when
   L2norm_fractional_tolerance==None are now gone.
   These code paths were apparently used by Mike for testing.
   There were two code paths: 1) Don't do any of the RPDMB operations
   and 2) bypass shuffle_widths, use the standard shuffling, and
   write different attributes into the h5 file.
   It appears to me that 2) would result in files that could not
   be read by rpdmb.load (since the attributes looked for by
   rpdmb.load would be missing).


sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
```diff
@@ -61,6 +62,8 @@ def save(
     L2norm_fractional_tolerance : float, optional
         Tolerance passed to `WaveformModes.truncate`; see that
         function's docstring for details.  Default value is 1e-10.
+        If L2norm_fractional_tolerance is <=0.0 or None, then no
+        truncation is done.
```
moble, line 66: We should just skip the `None` possibility.  The old code that would have used that is gone anyway, so it won't break anything more.  If anything, letting it succeed but take a different path would be a bad thing.
```suggestion
        If L2norm_fractional_tolerance is <=0.0, then no
        truncation is done.
```

sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
```diff
@@ -134,6 +138,13 @@ def save(
     if formats is None:
         formats = sxs_formats
 
+    if corotating_frame_tolerance is None:
+        corotating_frame_tolerance = L2norm_fractional_tolerance
+        if L2norm_fractional_tolerance is None or L2norm_fractional_tolerance <=0.0:
+            warning = f"corotating_frame_tolerance has not been specified, so it is defaulting to L2norm_fractional_tolerance.  However, L2norm_fractional_tolerance is {L2norm_fractional_tolerance} which does not make sense as a corotating_frame_tolerance.  You probably want to explicitly set corotating_frame_tolerance."
+            warnings.warn(warning)
+            
+
```
moble, line 147: The minimum sensible tolerance is 100*epsilon, so let's just fall back to that.

```suggestion
    if corotating_frame_tolerance is None:
        corotating_frame_tolerance = max(L2norm_fractional_tolerance, 100*np.finfo(float).eps)
```

sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
```diff
@@ -154,60 +165,57 @@ def save(
...
-            w.convert_to_conjugate_pairs()
+    # Convert mode structure to conjugate pairs
+    if convert_to_conjugate_pairs:
+        w.convert_to_conjugate_pairs()
 
-        # Set bits below the desired significance level to 0
-        if L2norm_fractional_tolerance > 0.0:
-            w.truncate(tol=L2norm_fractional_tolerance)
+    # Set bits below the desired significance level to 0
+    if L2norm_fractional_tolerance is not None and L2norm_fractional_tolerance > 0.0:
```
moble, line 203:
```suggestion
    if L2norm_fractional_tolerance > 0.0:
```

moble (changes requested):

PR #166, opened 2025-07-10
Modified align2d()

SunPeike:
1. Added an optional parameter `ftol` — tolerance for termination by the change of the cost function F. The optimization process is stopped when dF < ftol * F.
2. Added the trivial shift of `(δt, δϕ) = (0,0)`, and the trigger for that is` cost_wrapper([0.0, 0.0]) == 0`

-------  Optional to read ------
Initially, I thought `scipy.optimize.least_squares` would take several steps until the tolerance is met, then return the last step. And that is why we are missing `(0,0)`. To improve performance, I compared the cost of the last step and the previous one, choosing whichever was lower. Later, I realized this was unnecessary—`scipy.optimize.least_squares` interpolates between steps, so the final result may lie between them. As a result, comparing the last two steps wouldn't improve. Thus, just by simply check ` cost_wrapper([0.0, 0.0]) == 0` at the beginning is a simple and robust way

----
📚 Documentation preview 📚: https://sxs--166.org.readthedocs.build/en/166/

### DIFF
--- a/sxs/waveforms/alignment.py
+++ b/sxs/waveforms/alignment.py
@@ -197,10 +197,10 @@ def align2d(
     nprocs: int, optional
         Number of cpus to use. Default is maximum number.
         If -1 is provided, then no multiprocessing is performed.
-    max_nfev: None or int, optional
+    max_nfev: int, optional
         Parameter for scipy.optimize.least_squares.
         Controls the maximum number of function evaluations used. 
-        If None (default), the value is 100 * number of parameters
+        By default, max_nfev = 50000
     ftol: float, optional
         Parameter for scipy.optimize.least_squares.
         Tolerance for termination by the change of the cost function F. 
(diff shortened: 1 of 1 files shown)


sxs/waveforms/alignment.py
```diff
@@ -408,6 +427,14 @@ def align4d(
     nprocs: int, optional
         Number of cpus to use.  Default is maximum number.  If -1 is provided,
         then no multiprocessing is performed.
+    max_nfev: None or int, optional
+        Parameter for scipy.optimize.least_squares.
+        Controls the maximum number of function evaluations used. 
+        If None (default), the value is 100 * number of parameters
```
duetosymmetry, line 433: Docs claim default value is None, but on line 378 above, the default is specified as 50000

sxs/waveforms/alignment.py
```diff
@@ -509,7 +536,7 @@ def align4d(
 
     # Optimize by brute force with multiprocessing
     cost_wrapper = partial(_cost4d, args=[modes_A, modes_B, t_reference, normalization])
-
+
```
duetosymmetry, line 539: Remove whitespace change

sxs/waveforms/alignment.py
```diff
@@ -875,4 +902,4 @@ def align_simulations(
         nprocs=nprocs,
     )
 
-    return wa_prime, transformation, L2_norm, t1, t2
+    return wa_prime, transformation, L2_norm, t1, t2
```
duetosymmetry, line 905: Undo removing end of line before end of file

PR #190, opened 2026-03-23
Changing frame to take Q.array rather than np.array

akhairna:
Hi @moble and @duetosymmetry. I think this is a bug. The frame property should be of type `Q.array` rather than `np.array`. I found this while trying to load `sxs.WaveformModes` object to `scri.WaveformModes` object using `from_sxs` method in scri. Please let me know if this is an appropriate change.

----
📚 Documentation preview 📚: https://sxs--190.org.readthedocs.build/en/190/


akhairna:
Hi @moble, if this PR looks okay - can you merge it? The failure is due to an issue from PN.jl dependencies.

PR #191, opened 2026-03-31
Add pointwise optimized mismatch calculation

keefemitman:
This adds the usual LVK mismatch calculation of sampling a series of points on the two sphere, then for each point optimizing over time, binary phase, and polarization to minimize the residual, and returning the maximal mismatch obtained across all of the sky points. @moble probably worth you looking at, but @duetosymmetry probably also of interested to you.

----
📚 Documentation preview 📚: https://sxs--191.org.readthedocs.build/en/191/

### DIFF
--- a/sxs/waveforms/norms.py
+++ b/sxs/waveforms/norms.py
@@ -1,5 +1,4 @@
-import numpy as np
-import multiprocessing as mp
+import numpy as np import multiprocessing as mp
 from scipy.integrate import trapezoid
 from scipy.interpolate import CubicSpline
 from scipy.optimize import least_squares
@@ -302,8 +301,6 @@ def init_mismatch_pool(
     dphi_brute_force,
     limit_threads,
 ):
-    import sxs
-
     _MP_STATE.clear()
 
     if limit_threads:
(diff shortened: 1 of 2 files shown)


sxs/waveforms/norms.py
```diff
@@ -268,3 +264,470 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+    optimize_polarization,
+    dt_lower,
+    dt_upper,
+    x1,
+    x2,
+    dt_brute_force,
+    dphi_brute_force,
+    limit_threads,
+):
+    import sxs
```
duetosymmetry: Why is this import needed here?

keefemitman: Ah leftover from testing, whoops!

sxs/waveforms/norms.py
```diff
@@ -268,3 +264,470 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+    except Exception:
+        methods = []
+    return "fork" if "fork" in methods else "spawn"
+
+
+def limit_worker_threads():
+    try:
+        from threadpoolctl import threadpool_limits
+
+        return threadpool_limits(limits=1)
```
duetosymmetry, line 283: Do you want a new feature name under `[project.optional-dependencies]` in pyproject.toml so that threadpoolctl is an optional dep?

sxs/waveforms/norms.py
```diff
@@ -268,3 +264,470 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+        Beginning and end of integration interval.
+    N_theta : int, optional
+    N_phi : int, optional
+        Number of evenly spaced cos(theta) and phi values.
+        Default is 7 and 5.
+    optimize_polarization : bool, optional
+        Whether or not to choose an optimal polarization angle.
+        Default is True.
+    is_aligned : bool, optional
+        Whether or not to only consider points in the northern hemisphere.
```
duetosymmetry: Why does `is_aligned` mean only-northern? The name seems unintuitive

sxs/waveforms/norms.py
```diff
@@ -268,3 +264,470 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+
+    wa_sliced = wa[:, ia1:ia2]
+    wb_sliced = wb[:, ib1:ib2]
+
+    if include_modes is not None:
+        wa_sliced_data = wa_sliced.data
+        wb_sliced_data = wb_sliced.data
+        for mode in include_modes:
+            wa_sliced_data[:, wa_sliced.index(*mode)] *= 0
+            wb_sliced_data[:, wb_sliced.index(*mode)] *= 0
```
duetosymmetry: Isn't this the inverse of the mask you want? This is setting all modes in `include_modes` to vanish.

sxs/waveforms/norms.py
```diff
@@ -268,3 +264,470 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+    """
+    if t2 <= t1:
+        raise ValueError(f"(t1, t2)=({t1}, {t2}) is out of order.")
+    if wa.t[0] > t1 or wa.t[-1] < t2:
+        raise ValueError(f"(t1, t2)=({t1}, {t2}) is not contained in wa.t.")
+    if wb.t[0] > t1 or wb.t[-1] < t2:
+        raise ValueError(f"(t1, t2)=({t1}, {t2}) is not contained in wb.t.")
+    if wa.spin_weight != wb.spin_weight:
+        raise ValueError(
+            f"Spin weights must match, but got wa.spin_weight={wa.spin_weight} " f"and wb.spin_weight={wb.spin_weight}."
```
duetosymmetry:
```suggestion
            f"Spin weights must match, but got {wa.spin_weight=} and {wb.spin_weight=}."
```

sxs/waveforms/norms.py
keefemitman: Changing to `use_only_northern_hemisphere`

sxs/waveforms/norms.py
```diff
@@ -268,3 +264,469 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+    wb_sliced = wb[:, ib1:ib2]
+
+    if include_modes is not None:
+        wa_sliced_data = wa_sliced.data
+        wb_sliced_data = wb_sliced.data
+        for L in range(wa.ell_min, wa.ell_max + 1):
+            for M in range(-L, L + 1):
+                if not (L,M) in include_modes:
+                    wa_sliced_data[:, wa_sliced.index(L, M)] *= 0
+                    wb_sliced_data[:, wb_sliced.index(L, M)] *= 0
```
duetosymmetry, line 377: This is actually modifying the input, right? Is that intentional? The way it's called by `compute_optimized_mismatch_at_points` means that the arguments to `compute_optimized_mismatch_at_points` will themselves be modified.

sxs/waveforms/norms.py
```diff
@@ -268,3 +264,469 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
...
+
+    wa_sliced = wa[:, ia1:ia2].copy()
+    wb_sliced = wb[:, ib1:ib2].copy()
+
+    if include_modes is not None:
+        wa_sliced_data = wa_sliced.data
+        wb_sliced_data = wb_sliced.data
+        for L in range(wa.ell_min, wa.ell_max + 1):
+            for M in range(-L, L + 1):
+                if not (L,M) in include_modes:
```
duetosymmetry, line 375: I think stylistically we would prefer `(L,M) not in include_modes`, even though the result is the same

sxs/waveforms/norms.py
```diff
@@ -253,10 +252,7 @@ def mismatch(wa, wb, x1=-np.inf, x2=np.inf, modes=None, ASD=None):
                         wa.data[:, wa.index(L, M)] *= 0
                         wb.data[:, wb.index(L, M)] *= 0
         else:
-            raise ValueError(
-                "A `modes` argument was provided, but the input `wa` "
-                "and `wb` only have one dimension."
-            )
+            raise ValueError("A `modes` argument was provided, but the input `wa` " "and `wb` only have one dimension.")
```
duetosymmetry, line 255: This whitespace change looks funny...

PR #192, opened 2026-04-03
Updating Mma docs, including how to update WolframClientForPython

duetosymmetry:
----
📚 Documentation preview 📚: https://sxs--192.org.readthedocs.build/en/192/

### DIFF
--- a/docs/mathematica.md
+++ b/docs/mathematica.md
@@ -9,29 +9,74 @@ containing the Python code you want to run.
 
 As of this writing, Mathematica *claims* that it works with all reasonably
 modern versions of Python — specifically ["Python
-3.4+"](https://reference.wolfram.com/language/ref/externalevaluationsystem/Python.html).
-Assuming this is true, the only extra thing to worry about is that Mathematica
-requires the `pyzmq` package to interface with Python.
+3.5+"](https://reference.wolfram.com/language/ref/externalevaluationsystem/Python.html).
+The package `WolframClientForPython` ships with recent versions of Mathematica,
+and enables bidirectional communication between the two languages. The
+version of `WolframClientForPython` needs to be sufficiently new to avoid bugs
+in how they handle `pandas.NaT` (not-a-time) values that appear in our dataframe.
 
-If you have already installed `sxs`, you should be able to install `pyzmq` in
-the same way.  Alternatively, you may want to create a separate environment with
-both.  Using [`conda`](https://docs.anaconda.com/anaconda/install/), you can run
+/// details | `WolframClientForPython` newer than March 2026 required
+
+If the version of `WolframClientForPython` that shipped with your Mathematica
+installation is older than March 2026, then you have to install a newer version.
+At the time of writing, the newer version is not on PyPI, so you will have to
+install it directly from [their
+repo](https://github.com/WolframResearch/WolframClientForPython). Clone the
+repo, make sure you have all the requirements (including the `pyarrow` package),
+then do `pip install .` from the root of the `WolframClientForPython` directory,
+while you are in the appropriate mamba environment (see below).
+Next, we have "build" the paclet from within Mathematica so we can install it:
+```mathematica
+Needs["PacletTools`"]
+pbResult = PacletBuild["/path/to/local/WolframClientForPython"]
+```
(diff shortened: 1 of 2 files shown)


docs/mathematica.md
```diff
@@ -9,16 +9,59 @@ containing the Python code you want to run.
...
+and enables bidirectional communication between the two languages. The
+version of `WolframClientForPython` needs to be sufficiently new to avoid bugs
+in how they handle `pandas.NaT` (not-a-time) values that appear in our dataframe.
+
+/// details | `WolframClientForPython` newer than March 2026 required
+
+If the version of `WolframClientForPython` that shipped with your
+Mathematica installation is too old, then you have to install
+`WolframClientForPython` directly from [their
+repo](https://github.com/WolframResearch/WolframClientForPython). Clone the
```
moble, line 24: I'm confused about the version here.  Specifically, I immediately (before reading to the end of this selection) googled "WolframClientForPython" and found [this page](https://reference.wolfram.com/language/WolframClientForPython/), which says to `pip install wolframclient`, but that would appear to install [a version from June 2024](https://pypi.org/project/wolframclient/#history).  It might be worth linking to the relevant docs page [(this one?)](https://reference.wolfram.com/language/ref/externalevaluationsystem/Python.html) near where you introduce this, so people don't make the same mistake I did.  And maybe specifically say *not* to install via pip?

docs/mathematica.md
```diff
@@ -64,29 +107,29 @@ py["import sxs"];
...
-sims[
- Select[#["reference_mass_ratio"] > 8
-    && #["reference_dimensionless_spin1_mag"] < 0.1
-    && #["reference_dimensionless_spin2_mag"] < 0.1 &],
- {"initial_mass_ratio", "initial_separation", "reference_chi1_mag",
-  "reference_chi2_mag", "reference_eccentricity_bound"}]
+Select[#["reference_mass_ratio"] > 8
+     && #["reference_dimensionless_spin1_mag"] < 0.1
+     && #["reference_dimensionless_spin2_mag"] < 0.1 &][df][All,
+  {"Index1", "initial_mass_ratio", "initial_separation",
```
moble: What's the meaning of `"Index1"`?

docs/mathematica.md
duetosymmetry: Right, I poked them with a PR to update the version number and make a new PyPI release... they just don't really keep on top of this package. I will clarify. Of course it's possible that in the future we can tell people to just install from pip or conda.

docs/mathematica.md
duetosymmetry: This must be due to `WolframClientForPython` using pyarrow to ingest the pandas dataframe before serializing it. The row index (SXS ID) gets put into a new column named Index1 for some reason. If you want `Select[]` to keep the row index around, then you need to include `Index1` in the list

duetosymmetry: Ok I just dug a little deeper and it has nothing to do with pyarrow. `Index1` is automatically generated by pandas because we didn't explicitly specify the name of the index. Suppose in `sxscatalog/src/sxscatalog/simulations/simulations.py:Simulations.dataframe()`, after we make `sims_df`, we added the lines
```py
        # Set the name of the index to something meaningful
        sims_df.index.set_names("SXS ID", inplace=True)
```
then above, we would use `"SXS ID"` instead of `"Index1"`. I think this would simplify some things in the marimo notebook, too.

docs/mathematica.md
duetosymmetry: Updated

docs/mathematica.md
duetosymmetry: I can update this if sxs-collaboration/sxscatalog#13 is merged

duetosymmetry: Updated to use `"SXS ID"`

duetosymmetry:
Updated because Geoffrey noticed I used `sims` in one place but `df` in a different place in the Mma code.

Also realized that the main installation instructions say to use mamba rather than conda. @moble do you want the Mma docs to say to use mamba, as well? If yes, I will make that change in this PR, too.

moble:
Yeah, might as well; mamba is just a better user experience.  But maybe include some mention that it's basically the same thing as conda.

PR #193, opened 2026-04-09
load RIT and MAYA simulations dataframe

akhairna:
Hi @moble and @duetosymmetry. This small changes enables sxs package to load the RIT and MAYA simulations dataframe. Can you please review this?

----
📚 Documentation preview 📚: https://sxs--193.org.readthedocs.build/en/193/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ classifiers = [
   "Topic :: Scientific/Engineering :: Astronomy"
 ]
 dependencies = [
-  "sxscatalog >=3.0.14",
+  "sxscatalog >=3.0.23",
   "numpy >=2.0, <3.0",
   "scipy >=1.13",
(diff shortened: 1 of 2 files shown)


duetosymmetry:
The minimum version of `sxscatalog` in `pyproject.toml` needs to be updated

duetosymmetry:
This looks fine to me, you can squash. People won't really be able to use this until we also add the `RITSimulation` and `MAYASimulation` classes etc.... Not sure if we should try to do it all at the same time or bit by bit

akhairna:
Yes, I agree it needs more work for general use. However, I was currently using it to just obtain the metadata of simulations. So that I don't have to read my local catalog json file every time.

duetosymmetry (approved):

akhairna:
@moble - can you please merge this if there are no more comments?

PR #194, opened 2026-04-21
Add RITSimulation function and RITSimulation_v4 class

akhairna:
Hi @duetosymmetry and @moble. I have added a class and a function for loading the RIT waveforms. Using this `sxs.load` can load and cache RIT waveforms. Please review the PR.

----
📚 Documentation preview 📚: https://sxs--194.org.readthedocs.build/en/194/

### DIFF
--- a/sxs/__init__.py
+++ b/sxs/__init__.py
@@ -30,7 +30,7 @@ from .waveforms import rotating_paired_xor_multishuffle_bzip2 as rpxmb
 from .waveforms import rotating_paired_diff_multishuffle_bzip2 as rpdmb
 from .waveforms import spectre_cce_v1
 from . import catalog, metadata, horizons, waveforms, zenodo
-from .simulations import Simulation, Simulations, write_local_simulations, local_simulations
+from .simulations import Simulation, RITSimulation, Simulations, write_local_simulations, local_simulations
 from .handlers import load, load_via_sxs_id, loadcontext, load_lvc
 
 # The speed of light is, of course, defined to be exactly
--- a/sxs/__init__.py
+++ b/sxs/__init__.py
@@ -30,7 +30,10 @@ from .waveforms import rotating_paired_xor_multishuffle_bzip2 as rpxmb
 from .waveforms import rotating_paired_diff_multishuffle_bzip2 as rpdmb
 from .waveforms import spectre_cce_v1
 from . import catalog, metadata, horizons, waveforms, zenodo
-from .simulations import Simulation, RITSimulation, Simulations, write_local_simulations, local_simulations
+from .simulations import (
+    Simulation, RITSimulation, Simulations,
+    write_local_simulations, local_simulations,
+    )
 from .handlers import load, load_via_sxs_id, loadcontext, load_lvc
 
 # The speed of light is, of course, defined to be exactly
--- a/sxs/handlers.py
+++ b/sxs/handlers.py
@@ -234,6 +234,10 @@ def load(location, download=None, cache=None, progress=None, truepath=None, **kw
          the data.  Note that `download` must be explicitly set in
          this case, or a ValueError will be raised.
 
+      8) Given an RIT simulation specification — like "RIT:BBH:1234" or
+         "RIT:eBBH:1234" — the simulation is loaded as an
+         `sxs.RITSimulation` object.
+
     If the file is downloaded, it will be stored in the cache
     according to the `location`, unless `truepath` is set as noted
     above, in which case it is stored there.  Note that downloading is
@@ -243,7 +247,7 @@ def load(location, download=None, cache=None, progress=None, truepath=None, **kw
     """
     import pathlib
     import urllib.request
-    from . import Simulations, Simulation, read_config, sxs_directory, Catalog
+    from . import Simulations, Simulation, RITSimulation, read_config, sxs_directory, Catalog
     from .utilities import url, download_file, sxs_path_to_system_path, sxs_id_version_lev_exact_re, lev_path_re, sxs_identifier_re
 
     # Note: `download` and/or `cache` may still be `None` after this
@@ -296,6 +300,9 @@ def load(location, download=None, cache=None, progress=None, truepath=None, **kw
 
         elif sxs_id_version_lev_exact_re.match(location):
             return Simulation(location, download=download, cache=cache, progress=progress, **kwargs)
+
+        elif location.startswith("RIT:"):
+            return RITSimulation(location, download=download, cache=cache, progress=progress, **kwargs)
 
         else:
             # Try to find an appropriate SXS file in the simulations
--- a/sxs/handlers.py
+++ b/sxs/handlers.py
@@ -247,7 +247,10 @@ def load(location, download=None, cache=None, progress=None, truepath=None, **kw
     """
     import pathlib
     import urllib.request
-    from . import Simulations, Simulation, RITSimulation, read_config, sxs_directory, Catalog
+    from . import (
+        Simulations, Simulation, RITSimulation,
+        read_config, sxs_directory, Catalog
+    )
     from .utilities import url, download_file, sxs_path_to_system_path, sxs_id_version_lev_exact_re, lev_path_re, sxs_identifier_re
 
     # Note: `download` and/or `cache` may still be `None` after this
--- a/sxs/simulations/rit_simulation.py
+++ b/sxs/simulations/rit_simulation.py
@@ -1,8 +1,8 @@
 import urllib
-from pathlib import PurePosixPath
+from pathlib import PurePosixPath, Path
 from .simulation import SimulationBase
 from .. import Metadata
-from ..utilities import download_file, sxs_directory
+from ..utilities import download_file, sxs_directory, sxs_path_to_system_path
 
 
 def _not_defined_property(name):
@@ -70,7 +70,7 @@ def RITSimulation(location, *args, **kwargs):
         location,
         *args, **kwargs
     )
-    sim.__file__ = str(sxs_directory("cache") / location)
+    sim.__file__ = str(sxs_directory("cache") / sxs_path_to_system_path(location))
     return sim
 
 
@@ -106,37 +106,40 @@ class RITSimulation_v4(SimulationBase):
         construction += f"e={e:.3g} simulation" if type(e) is float else f"{e=} simulation"
         return construction
 
-    def load_waveform(self, location):
-
-        from ..waveforms import lvcnr
-
-        strain_url = self.metadata.extrap_strain_url
-        strain_path = urllib.parse.urlparse(strain_url).path
-        filename = PurePosixPath(strain_path).name
-
-        path = sxs_directory("cache") / location / filename
-
-        if not path.exists():
-            download_file(strain_url, path)
-
-        w = lvcnr.load(path)
-        w.metadata = self.metadata
-
-        return w
-
     distances = _not_implemented_function("distances")
     closest_simulation = _not_implemented_function("closest_simulation")
+    load_waveform = _not_implemented_function("load_waveform")
     load_horizons = _not_implemented_function("load_horizons")
 
     @property
     def strain(self):
         if not hasattr(self, "_strain"):
-            self._strain = self.load_waveform(self.location)
+            from ..waveforms import lvcnr
+
+            strain_url = self.metadata.extrap_strain_url
+            strain_path = urllib.parse.urlparse(strain_url).path
+            filename = PurePosixPath(strain_path).name
+            location = Path(self.location)
+
+            path = sxs_directory("cache") / location / filename
+
+            if not path.exists():
+                download_file(strain_url, path)
+
+            w = lvcnr.load(path)
+            w.metadata = self.metadata
+
+            self._strain = w
         return self._strain
 
     Strain = strain
     h = strain
     H = strain
+
+    @property
+    def psi4(self):
+        raise NotImplementedError(f"psi4 is currently not implemented for RITSimulation_v4.")
+    Psi4 = psi4
 
     versions = _not_defined_property("versions")
     lev = _not_defined_property("lev")
--- a/sxs/simulations/rit_simulation.py
+++ b/sxs/simulations/rit_simulation.py
@@ -1,4 +1,5 @@
-from pathlib import Path
+import urllib
+from pathlib import PurePosixPath
 from .simulation import SimulationBase
 from .. import Metadata
 from ..utilities import download_file, sxs_directory
@@ -69,6 +70,7 @@ def RITSimulation(location, *args, **kwargs):
         version=version,
         files=files,
         location=location,
+        *args, **kwargs
     )
     sim.__file__ = str(sxs_directory("cache") / location)
     return sim
@@ -111,7 +113,8 @@ class RITSimulation_v4(SimulationBase):
         from ..waveforms import lvcnr
 
         strain_url = self.metadata.extrap_strain_url
-        filename = Path(strain_url).name
+        strain_path = urllib.parse.urlparse(strain_url).path
+        filename = PurePosixPath(strain_path).name
 
         path = sxs_directory("cache") / location / filename
 
--- a/sxs/simulations/rit_simulation.py
+++ b/sxs/simulations/rit_simulation.py
@@ -8,8 +8,7 @@ from ..utilities import download_file, sxs_directory
 def _not_defined_property(name):
     @property
     def prop(self):
-        raise AttributeError(f"'{name}' is not supported for version {self.version} of the catalog.")
-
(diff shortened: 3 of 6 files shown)


sxs/simulations/ritsimulation.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def RITSimulation(location, *args, **kwargs):
+    """Construct a RITSimulation object from a location string.
+
+    The location string should be an RIT ID, as in "RIT:BBH:xxxx" or
+    "RIT:eBBH:xxxx". It must exist, or no files will be found to load the data
+    from.
+
+    The returned object will be either a `RITSimulation_v4` object.
```
duetosymmetry: `either` is copy/pasted from SXS, but right now we only have one version of RIT

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,151 @@
...
+class RITSimulation_v4(SimulationBase):
+    """Simulation object for version 4 of the RIT data format
+
+    Note that users almost certainly never need to call this function;
+    see the `RITSimulation` function or `sxs.load` function instead.  See
+    also `SimulationBase` for the base class that this class inherits
+    from.
+    """
+
+    def __init__(self, metadata, series, version, files, location, *args, **kwargs):
```
duetosymmetry, line 86: Usually you would call `super().__init__(...)`. It will actually set all of these for you, so you don't have to repeat the below code.

sxs/simulations/ritsimulation.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Parameters
+    ----------
+    location : str
+        The location string for the simulation.  This must include the
+        RIT ID as described above.
+
+    Keyword Arguments
+    -----------------
+    download_file_info : bool, optional
```
duetosymmetry: `download_file_info` is never checked

duetosymmetry:
BTW I think typical python naming would suggest to name the file `rit_simulation.py` instead of `ritsimulation.py`

duetosymmetry:
Is there a test we can add?

sxs/simulations/rit_simulation.py
akhairna: Yes, I was using super().__init__ initially. But that had extra positional arguments like `lev`, `lev_numbers` etc which aren't valid for RIT. Instead of setting them to `None`, I thought this is more clear way to write the constructor.

duetosymmetry: Good point; then I'd say to include a comment explaining why you're not calling `super().__init__(..)`

sxs/simulations/ritsimulation.py
akhairna: Oops missed that one. Will remove that part from the docstring. It is sxs specific. That's why it can't be used later.

akhairna:
> Is there a test we can add?

I will add some test in `test_simulation.py`.

sxs/handlers.py
```diff
@@ -297,6 +301,9 @@ def load(location, download=None, cache=None, progress=None, truepath=None, **kw
         elif sxs_id_version_lev_exact_re.match(location):
             return Simulation(location, download=download, cache=cache, progress=progress, **kwargs)
 
+        elif location.startswith("RIT:"):
+            return RITSimulation(location)
```
duetosymmetry: For future flexibility, you want to pass `**kwargs`. It doesn't hurt to pass all the arguments (download, cache, and progress) just in case we want to actually do something with them.

akhairna: okay, will add that.

akhairna:
I rebased my branch on top of the main. Also, I added a test in `test_simulation.py`. Let me know if you think there is any other test that can be added. Other kind of tests were specific to sxs simulations, that's why I did those.

tests/test_simulation.py
```diff
@@ -290,3 +290,14 @@ def test_issue_116():
...
+
+@skip_macOS_GH_actions_downloads
+def test_sxs_load_rit_v4():
+    s = sxs.load("RIT:BBH:0084")
+    assert s.location == "RIT:BBH:0084"
+    assert s.version == "v4.0"
+    s.h
+    s.strain
+    s.H
+    s.Strain
```
duetosymmetry, line 303: The `test_sxs_load*` functions also check the time at the end with `assert s.h.t[0] >= 0`.
Is there anything to test aside from the fact that downloading actually works?

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,144 @@
...
+        RIT ID as described above.
+
+    Returns
+    -------
+    simulation : SimulationBase
+        A `RITSimulation_v4` object, depending on the version of the simulation
+        data.
+
+    Note that all remaining arguments (including keyword arguments)
+    are passed on to the `SimulationBase` constructors.
```
duetosymmetry, line 43: Additional `args` and `kwargs` don't actually seem to be passed to anywhere. Do you want to include them when creating `sim = RITSimulation_v4(...)` below?

sxs/__init__.py
```diff
@@ -30,7 +30,7 @@
 from .waveforms import rotating_paired_diff_multishuffle_bzip2 as rpdmb
 from .waveforms import spectre_cce_v1
 from . import catalog, metadata, horizons, waveforms, zenodo
-from .simulations import Simulation, Simulations, write_local_simulations, local_simulations
+from .simulations import Simulation, RITSimulation, Simulations, write_local_simulations, local_simulations
```
duetosymmetry: IMO this line is long enough that I would break it:
```suggestion
from .simulations import (
    Simulation, RITSimulation, Simulations,
    write_local_simulations, local_simulations,
)
```

sxs/handlers.py
```diff
@@ -243,7 +247,7 @@ def load(location, download=None, cache=None, progress=None, truepath=None, **kw
     """
     import pathlib
     import urllib.request
-    from . import Simulations, Simulation, read_config, sxs_directory, Catalog
+    from . import Simulations, Simulation, RITSimulation, read_config, sxs_directory, Catalog
```
duetosymmetry: Same here with the long line

sxs/simulations/rit_simulation.py
akhairna: Added this.

tests/test_simulation.py
akhairna: I didn't add this because the RIT strain data have aligned their t=0 with the merger. Therefore s.h.t[0] would be less than 0 at all times.

sxs/handlers.py
akhairna: I didn't run black which would have fixed this because it changes a lot of things. Nonetheless I fixed this one line manually.

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,145 @@
...
+        construction += f"""chi2=[{", ".join(f"{c:.3g}" for c in chi2)}] """
+        construction += f"e={e:.3g} simulation" if type(e) is float else f"{e=} simulation"
+        return construction
+
+    def load_waveform(self, location):
+
+        from ..waveforms import lvcnr
+
+        strain_url = self.metadata.extrap_strain_url
+        filename = Path(strain_url).name
```
duetosymmetry: I think this is an unexpected use of `Path()`. It's intended to give paths to files on disk, not parse URLs. I think we would normally use `urllib.parse.urlparse(url).path` (which you could then split to get the file name).

tests/test_simulation.py
duetosymmetry: I didn't mean specifically the fact that the first time is non-negative, but just doing something with the values at all

akhairna: Added a small assertion.

duetosymmetry: It looks a bit funny to me to do exact comparison with 0 for a floating point number that comes from a nontrivial calculation... I would have done `np.allclose()`, but since it's only for one specific waveform and it the test passes, I'm not going to worry a lot.

duetosymmetry (approved):
I'm happy with this! Can be squashed.

akhairna:
@moble - let me know if this is okay. I have squashed the 6 commits down to 2. I can squash it down further if needed. Otherwise it is ready.

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+        e = self.series.eccentricity
+        construction = f"""{type(self).__qualname__}("{self.location}")\n# """
+        construction += f"n_orbits={self.metadata.number_of_orbits:.3g} "
+        construction += f"q={(1/self.metadata.relaxed_mass_ratio_1_over_2):.3g} "
+        construction += f"""chi1=[{", ".join(f"{c:.3g}" for c in chi1)}] """
+        construction += f"""chi2=[{", ".join(f"{c:.3g}" for c in chi2)}] """
+        construction += f"e={e:.3g} simulation" if type(e) is float else f"{e=} simulation"
+        return construction
+
+    def load_waveform(self, location):
```
moble: This function is overriding the function from the base class, which has a different call signature.  That means this function will just produce incoherent errors if called from other code.  It would be better to either handle the same arguments as the base class, or move the logic that's currently here into the `strain` function and just raise an error in this function saying that that function is not implemented for RITSimulation_v4.

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+    @property
+    def strain(self):
+        if not hasattr(self, "_strain"):
+            self._strain = self.load_waveform(self.location)
+        return self._strain
+
+    Strain = strain
+    h = strain
+    H = strain
+
```
moble, line 138: `SimulationBase.psi4` is inherited by this class, and will call `self.load_waveform(*self.psi4_path)`, which will error.  The Psi4 data appears (from the `files` variable in `RITSimulation`) to be available; why not support it too?  If not, it should be a property that raises an explanatory error.

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+    versions = _not_defined_property("versions")
+    lev = _not_defined_property("lev")
+    Lev = _not_defined_property("Lev")
+    metadata_path = _not_defined_property("metadata_path")
+    horizons = _not_defined_property("horizons")
```
moble, line 148: I believe `Horizons` would resolve to `SimulationBase.Horizons`, which would fail.
```suggestion
    horizons = _not_defined_property("horizons")
    Horizons = horizons
```

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+    see the `RITSimulation` function or `sxs.load` function instead.  See
+    also `SimulationBase` for the base class that this class inherits
+    from.
+    """
+
+    def __init__(self, metadata, series, version, files, location, *args, **kwargs):
+    # Note that we want just a subset of the constructor of SimulationBase here.
+    # Therefore, we don't use the super of SimulationsBase.
+    # This keeps the constructor clean and avoids setting positional arguments -
+    # sxs_id_stem, sxs_id, url, lev_numbers and lev_number to None.
```
moble:
```suggestion
        # Note that we want just a subset of the constructor of SimulationBase here.
        # Therefore, we don't use the super of SimulationsBase.
        # This keeps the constructor clean and avoids setting positional arguments -
        # sxs_id_stem, sxs_id, url, lev_numbers and lev_number to None.
```

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+    version = "v4.0"
+
+    sim = RITSimulation_v4(
+        metadata=metadata,
+        series=series,
+        version=version,
+        files=files,
+        location=location,
+        *args, **kwargs
+    )
```
moble, line 72: These aren't kwargs, and kwargs shouldn't be followed by `*args` anyway.
```suggestion
    sim = RITSimulation_v4(
        metadata,
        series,
        version,
        files,
        location,
        *args, **kwargs
    )
```

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+from .simulation import SimulationBase
+from .. import Metadata
+from ..utilities import download_file, sxs_directory
+
+
+def _not_defined_property(name):
+    @property
+    def prop(self):
+        raise AttributeError(f"'{name}' is not supported for version {self.version} of the catalog.")
+
```
moble: "Supported" makes it sound like we just haven't gotten around to any of these features.  Also, the "RIT" part of this error should be made explicit.
```suggestion
    def prop(self):
        raise AttributeError(f"'{name}' is not available for RIT simulations (version {self.version}).")
```

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+        chi1 = self.series["relaxed_chi1"]
+        chi2 = self.series["relaxed_chi2"]
+        e = self.series.eccentricity
+        construction = f"""{type(self).__qualname__}("{self.location}")\n# """
+        construction += f"n_orbits={self.metadata.number_of_orbits:.3g} "
+        construction += f"q={(1/self.metadata.relaxed_mass_ratio_1_over_2):.3g} "
+        construction += f"""chi1=[{", ".join(f"{c:.3g}" for c in chi1)}] """
+        construction += f"""chi2=[{", ".join(f"{c:.3g}" for c in chi2)}] """
+        construction += f"e={e:.3g} simulation" if type(e) is float else f"{e=} simulation"
+        return construction
```
moble, line 107: Are we really sure these fields will always exist?  It can be really annoying to try to debug code when the issue is that the display of the returned object is the part that's failing.  If you're not 100% sure, it would be better to try to guard these things with `get` and a default.

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+    ----------
+    location : str
+        The location string for the simulation.  This must include the
+        RIT ID as described above.
+
+    Returns
+    -------
+    simulation : SimulationBase
+        A `RITSimulation_v4` object, depending on the version of the simulation
+        data.
```
moble:
```suggestion
        An `RITSimulation_v4` object.
```

sxs/simulations/rit_simulation.py
```diff
@@ -0,0 +1,147 @@
...
+
+    sim = RITSimulation_v4(
+        metadata=metadata,
+        series=series,
+        version=version,
+        files=files,
+        location=location,
+        *args, **kwargs
+    )
+    sim.__file__ = str(sxs_directory("cache") / location)
```
moble: `location` needs to be wrapped in `sxs_path_to_system_path`, which needs to be imported from utilities.

moble (changes requested):
We should probably bump the `sxscatalog` dependency in `pyproject.toml` to 3.0.28 to catch the bug fixes related to RITSimulations.

moble:
> @moble - let me know if this is okay. I have squashed the 6 commits down to 2. I can squash it down further if needed. Otherwise it is ready.

Please never squash anything once anyone has commented on a PR.  I'll be happy to squash if I want to, but force pushes just screw up the github interface and leads to lost comments.

sxs/simulations/rit_simulation.py
akhairna: Fixed this.

sxs/simulations/rit_simulation.py
akhairna: Yes Psi4 data is available from RIT. However, I decided to add the attribute later mainly due to urgency of my work that needed only strain. I will add an explanatory error.

sxs/simulations/rit_simulation.py
akhairna: The current arguments of `load_waveform` were tied to SXS simulations that's why I changed it. But I didn't realize the errors that it may produce. I will add a not implemented error for this function and move the logic to strain as you suggested. Maybe later I can include this functionality.

sxs/simulations/rit_simulation.py
akhairna: These fields are read from the contents of the json file in `sxscatalog`. I think @duetosymmetry added `nan` values for fields that aren't present.

akhairna:
@moble & @duetosymmetry - please review these changes. Let me know what you think.

akhairna:
@moble - please let me know if these changes are ready to merge!

sxs/simulations/rit_simulation.py
duetosymmetry: @akhairna is deriving the `series` object from the `.dataframe` attribute; and for the dataframe, we always have all of these fields. They are filled in with some values (either 0's or NaNs, depending on what is appropriate) if they are missing.

Issue #21, opened 2020-09-21
Unify treatment of Modes objects


moble:
We now have `spherical.Modes`, `scri.ModesTimeSeries`, `scri.AsymptoticBondiData`, `sxs.TimeSeries`, and `sxs.WaveformModes` objects — with sporadic relationships between them.  I'm thinking that `scri.ModesTimeSeries` should be moved to `sxs`, and inherit from both `spherical.Modes` and `sxs.TimeSeries`.  Then, `sxs.WaveformModes` can inherit from the new `sxs.ModesTimeSeries` and `WaveformMixin`, while `sxs.WaveformGrid` can inherit from `sxs.TimeSeries` and `WaveformMixin`.

PR #195, opened 2026-05-12
Add mul method to wmodes

akhairna:
Hi @moble, @duetosymmetry, and @keefemitman. I am adding the `__mul__` method to the waveform modes class. Using this we can implement multiplication between waveform modes of different array size. I am using the attributes of the first Wmodes object to define the `l_max` of product. Please suggest improvements for this.

----
📚 Documentation preview 📚: https://sxs--195.org.readthedocs.build/en/195/

### DIFF
--- a/sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
+++ b/sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
@@ -567,6 +567,7 @@ def load(
         ell_min=ell_min,
         ell_max=ell_max,
         spin_weight=spin_weight,
+        multiplication_truncator=max
     )
 
     if convert_from_conjugate_pairs:
--- a/sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
+++ b/sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
@@ -567,7 +567,6 @@ def load(
         ell_min=ell_min,
         ell_max=ell_max,
         spin_weight=spin_weight,
-        multiplication_truncator=max
     )
 
     if convert_from_conjugate_pairs:
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -150,6 +150,33 @@ class WaveformModes(WaveformMixin, TimeSeries):
                 obj = TimeSeries(obj)
         return obj
 
+    def __mul__(self, other):
+        if isinstance(other, WaveformModes):
+            modes12_data, modes12_ellmin, modes12_ellmax, modes12_spin = spherical.multiply(
+                self,
+                self.ell_min,
+                self.ell_max,
+                self.spin_weight,
+                other,
+                other.ell_min,
+                other.ell_max,
+                other.spin_weight,
+                ellmin_fg=None,
+                ellmax_fg=self.ell_max
+            )
+            return WaveformModes(
+                modes12_data,
+                time=self.time,
+                time_axis=0,
+                ell_min=modes12_ellmin,
+                ell_max=modes12_ellmax,
+                modes_axis=1,
+                spin_weight=modes12_spin,
+                frame=self.frame
+            )
+        else:
+            return super().__mul__(other)
+
     @property
     def modes_axis(self):
         """Axis of the array storing the various modes
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -154,7 +154,7 @@ class WaveformModes(WaveformMixin, TimeSeries):
         """Addition method for two WaveformModes object.
 
         Raises TypeError if other is not a WaveformModes instance.
-        Raises ValueError if spin weights or data shapes are incompatible.
+        Raises ValueError if spin weights are incompatible.
         """
         if not isinstance(other, type(self)):
             raise TypeError(f"Cannot add WaveformModes class object with an object of type {type(other).__name__}.")
@@ -163,18 +163,33 @@ class WaveformModes(WaveformMixin, TimeSeries):
 
         if self.spin_weight != other.spin_weight:
             raise ValueError(f"Cannot add two WaveformModes with different spin weights ({self.spin_weight=} and {other.spin_weight=})")
-        if self.shape != other.shape:
-            raise ValueError(f"Cannot add two WaveformModes with different shape of the data array. ({self.shape=} and {other.shape=})")
 
-        result = self.data + other.data
+        mode_diff = self.n_modes - other.n_modes
 
-        return type(self)(result, **self._metadata)
+        data_self = np.pad(self.data, pad_width=((0, 0), (-mode_diff, 0))) if mode_diff < 0 else self.data
+
+        data_other = np.pad(other.data, pad_width=((0,0),(mode_diff,0))) if mode_diff > 0 else other.data
+
+        result = data_self + data_other
+        ell_min = min(self.ell_min, other.ell_min)
+
+        return type(self)(
+            result,
+            time=self.time,
+            time_axis=0,
+            ell_min=ell_min,
+            ell_max=self.ell_max,
+            modes_axis=1,
+            spin_weight=self.spin_weight,
+            frame=self.frame,
+            frame_type=self.frame_type,
+        )
 
     def __sub__(self, other):
         """Subtraction method for two WaveformModes object.
 
         Raises TypeError if other is not a WaveformModes instance.
-        Raises ValueError if spin weights or data shapes are incompatible.
+        Raises ValueError if spin weights are incompatible.
         """
         if not isinstance(other, type(self)):
             raise TypeError(f"Cannot subtract WaveformModes class object with an object of type {type(other).__name__}.")
@@ -183,12 +198,27 @@ class WaveformModes(WaveformMixin, TimeSeries):
 
         if self.spin_weight != other.spin_weight:
             raise ValueError(f"Cannot subtract two WaveformModes with different spin weights ({self.spin_weight=} and {other.spin_weight=})")
-        if self.shape != other.shape:
-            raise ValueError(f"Cannot subtract two WaveformModes with different shape of the data array. ({self.shape=} and {other.shape=})")
 
-        result = self.data - other.data
+        mode_diff = self.n_modes - other.n_modes
 
-        return type(self)(result, **self._metadata)
+        data_self = np.pad(self.data, pad_width=((0, 0), (-mode_diff, 0))) if mode_diff < 0 else self.data
+
+        data_other = np.pad(other.data, pad_width=((0,0),(mode_diff,0))) if mode_diff > 0 else other.data
+
+        result = data_self - data_other
+        ell_min = min(self.ell_min, other.ell_min)
+
+        return type(self)(
+            result,
+            time=self.time,
+            time_axis=0,
+            ell_min=ell_min,
+            ell_max=self.ell_max,
+            modes_axis=1,
+            spin_weight=self.spin_weight,
+            frame=self.frame,
+            frame_type=self.frame_type,
+        )
 
     def __mul__(self, other):
         """Multiplication method for two WaveformModes object.
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -151,19 +151,28 @@ class WaveformModes(WaveformMixin, TimeSeries):
         return obj
 
     def __mul__(self, other):
-        if isinstance(other, WaveformModes):
+        if isinstance(other, WaveformModes) and np.allclose(self.frame, other.frame):
+            if self.multiplication_truncator is not None:
+                new_ell_max = self.multiplication_truncator((self.ell_max, other.ell_max))
+            else:
+                new_ell_max = max(
+                    truncator((self.ell_max, other.ell_max))
+                    for truncator in [
+                        self._metadata.get('multiplication_truncator', sum),
+                        other._metadata.get('multiplication_truncator', sum)
+                    ]
+                )
             modes12_data, modes12_ellmin, modes12_ellmax, modes12_spin = spherical.multiply(
-                self,
-                self.ell_min,
-                self.ell_max,
-                self.spin_weight,
-                other,
-                other.ell_min,
-                other.ell_max,
-                other.spin_weight,
-                ellmin_fg=None,
-                ellmax_fg=self.ell_max
-            )
+            self,
+            self.ell_min,
+            self.ell_max,
+            self.spin_weight,
+            other,
+            other.ell_min,
+            other.ell_max,
+            other.spin_weight,
+            ellmax_fg=new_ell_max
+        )
             return WaveformModes(
                 modes12_data,
                 time=self.time,
@@ -199,6 +208,11 @@ class WaveformModes(WaveformMixin, TimeSeries):
     def ell_max(self):
         """Largest value of ℓ stored in the data"""
         return self._metadata["ell_max"]
+
+    @property
+    def multiplication_truncator(self):
+        """Multiplication truncator stored in the data"""
+        return self._metadata["multiplication_truncator"]
 
     @property
     def n_modes(self):
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -220,14 +220,16 @@ class WaveformModes(WaveformMixin, TimeSeries):
         return self + (-other)
 
     def __mul__(self, other):
-        """Multiplication method for two WaveformModes object.
+        """Multiplication method for two WaveformModes object; falls back to
+        the parent class if `other` is not a WaveformModes instance.
 
         First it checks if the two objects are compatible for multiplication
-        using the `_require_compatibility` method.
-        The output `ell_max` is determined by the `multiplication_truncator`
-        attribute. The left operand's truncator takes priority; if absent, the
-        right operand's truncator is used; if both are absent, it defaults to
-        the sum of the two `ell_max`.
+        using the `_require_compatibility` method and `modes_axis` of the
+        data.
+        The product `ell_max` and `multiplication_truncator` are determined by
+        applying each operand's `multiplication_truncator` to `(self.ell_max,
+        other.ell_max)`, and selecting the operand whose truncator yields the
+        larger result.
         ----
         Note that the `modes_axis` of self and other should be the last axis of
         the WaveformModes data as per spherical.multiply, otherwise a
@@ -236,12 +238,12 @@ class WaveformModes(WaveformMixin, TimeSeries):
         if not isinstance(other, type(self)):
             return super().__mul__(other)
 
+        self._require_compatibility(other)
+
         if self.modes_axis != self.ndim - 1:
             raise ValueError(
                 f"The modes_axis of self and other should be the last axis of the data (ndim-1={self.ndim - 1}), rather it is {self.modes_axis}."
             )
-
-        self._require_compatibility(other)
 
         self_ell_max = self.multiplication_truncator((self.ell_max, other.ell_max))
         other_ell_max = other.multiplication_truncator((self.ell_max, other.ell_max))
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -338,7 +338,7 @@ class WaveformModes(WaveformMixin, TimeSeries):
         return self._metadata.get("multiplication_truncator", max)
 
     @multiplication_truncator.setter
-    def multiplication_truncation(self, truncator_prm):
+    def multiplication_truncator(self, truncator_prm):
         self._metadata["multiplication_truncator"] = truncator_prm
 
     @property
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -164,21 +164,24 @@ class WaveformModes(WaveformMixin, TimeSeries):
         if self.spin_weight != other.spin_weight:
             raise ValueError(f"Cannot add two WaveformModes with different spin weights ({self.spin_weight=} and {other.spin_weight=})")
 
-        mode_diff = self.n_modes - other.n_modes
-
-        data_self = np.pad(self.data, pad_width=((0, 0), (-mode_diff, 0))) if mode_diff < 0 else self.data
-
-        data_other = np.pad(other.data, pad_width=((0,0),(mode_diff,0))) if mode_diff > 0 else other.data
-
-        result = data_self + data_other
         ell_min = min(self.ell_min, other.ell_min)
+        ell_max = max(self.ell_max, other.ell_max)
+
+        result = np.zeros((self.time.size, spherical.Ysize(ell_min, ell_max)), dtype=self.dtype)
+
+        slice_1 = slice(spherical.Yindex(self.ell_min, -self.ell_min, ell_min), spherical.Yindex(self.ell_max + 1, -(self.ell_max + 1), ell_min),)
+
+        slice_2 = slice(spherical.Yindex(other.ell_min, -other.ell_min, ell_min), spherical.Yindex(other.ell_max + 1, -(other.ell_max + 1), ell_min),)
+
+        result[:, slice_1] += self.data
+        result[:, slice_2] += other.data
 
         return type(self)(
             result,
             time=self.time,
             time_axis=0,
             ell_min=ell_min,
-            ell_max=self.ell_max,
+            ell_max=ell_max,
             modes_axis=1,
             spin_weight=self.spin_weight,
             frame=self.frame,
@@ -199,21 +202,24 @@ class WaveformModes(WaveformMixin, TimeSeries):
         if self.spin_weight != other.spin_weight:
             raise ValueError(f"Cannot subtract two WaveformModes with different spin weights ({self.spin_weight=} and {other.spin_weight=})")
 
-        mode_diff = self.n_modes - other.n_modes
-
-        data_self = np.pad(self.data, pad_width=((0, 0), (-mode_diff, 0))) if mode_diff < 0 else self.data
-
-        data_other = np.pad(other.data, pad_width=((0,0),(mode_diff,0))) if mode_diff > 0 else other.data
-
-        result = data_self - data_other
         ell_min = min(self.ell_min, other.ell_min)
+        ell_max = max(self.ell_max, other.ell_max)
+
+        result = np.zeros((self.time.size, spherical.Ysize(ell_min, ell_max)), dtype=self.dtype)
+
+        slice_1 = slice(spherical.Yindex(self.ell_min, -self.ell_min, ell_min), spherical.Yindex(self.ell_max + 1, -(self.ell_max + 1), ell_min),)
+
+        slice_2 = slice(spherical.Yindex(other.ell_min, -other.ell_min, ell_min), spherical.Yindex(other.ell_max + 1, -(other.ell_max + 1), ell_min),)
+
+        result[:, slice_1] += self.data
+        result[:, slice_2] -= other.data
 
         return type(self)(
             result,
             time=self.time,
             time_axis=0,
             ell_min=ell_min,
-            ell_max=self.ell_max,
+            ell_max=ell_max,
             modes_axis=1,
             spin_weight=self.spin_weight,
             frame=self.frame,
@@ -252,7 +258,8 @@ class WaveformModes(WaveformMixin, TimeSeries):
         other.ell_max,
         other.spin_weight,
         ellmax_fg=new_ell_max
-    )
+        )
+
         ell_min = abs(modes12_spin)
         modes_data = modes12_data[:,spherical.Yindex(ell_min, - ell_min, modes12_ellmin) : ]
 
@@ -286,6 +293,8 @@ class WaveformModes(WaveformMixin, TimeSeries):
             raise ValueError(f"Both waveforms must have identical frame arrays.")
         if not np.array_equal(self.time, other.time):
             raise ValueError(f"Both waveforms must have identical time arrays.")
+        if not self.time_axis==other.time_axis:
+            raise ValueError(f"The time axis of the two waveforms must be same for algebraic operations.")
 
         return True
 
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -150,6 +150,46 @@ class WaveformModes(WaveformMixin, TimeSeries):
                 obj = TimeSeries(obj)
         return obj
 
+    def __add__(self, other):
+        """Addition method for two WaveformModes object.
+
+        Raises TypeError if other is not a WaveformModes instance.
+        Raises ValueError if spin weights or data shapes are incompatible.
+        """
+        if not isinstance(other, WaveformModes):
+            raise TypeError(f"Cannot add WaveformModes class object with an object of type {type(other).__name__}.")
+
+        self._is_compatible(other)
+
+        if self.spin_weight != other.spin_weight:
+            raise ValueError(f"Cannot add two WaveformModes with different spin weights ({self.spin_weight=} and {other.spin_weight=})")
+        if self.shape != other.shape:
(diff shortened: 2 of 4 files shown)


sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,33 @@ def __getitem__(self, key):
...
+                self,
+                self.ell_min,
+                self.ell_max,
+                self.spin_weight,
+                other,
+                other.ell_min,
+                other.ell_max,
+                other.spin_weight,
+                ellmin_fg=None,
+                ellmax_fg=self.ell_max
```
keefemitman: I think we should use the minimum of the two `ell_max`'s?

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,33 @@ def __getitem__(self, key):
                 obj = TimeSeries(obj)
         return obj
 
+    def __mul__(self, other):
+        if isinstance(other, WaveformModes):
```
keefemitman: We should probably require that the frames are the same too, no?

keefemitman (changes requested):
Thanks @akhairna! Just two minor comments/edit requests.

sxs/waveforms/waveform_modes.py
akhairna: Hmm, I was wondering about that too. Whether minimum or maximum of the two? Also, it doesn't matter significantly for the output of CCE because the `lmax` is set to be 8 (or same) for all fields (I believe this is the case.)

After creating this PR, I realized that the issue is coming primarily from the `ellmin_fg`. If it is set to None, then it will take ell_min(f,g) as per PR moble/spherical#28. If I don't set the minimum to 0, this gives me error while testing the constraints like the Bianchi identity 0.

duetosymmetry: I think it's reasonable to have some default, but I think it's overly restrictive to just pick one of those options and put it in. The approach in `scri` was to have an attribute, `multiplication_truncator`, which was passed to the appropriate `spherical_functions` method. That way you can change to e.g. (i) not losing any information by using the sum; or (ii) keeping a certain max ell by using max; or maybe (iii) something else that's reasonable?

akhairna: Yes, I can do that too. Whatever we think is the most reasonable.

sxs/waveforms/waveform_modes.py
akhairna: Yes, I can add a check.

akhairna:
@duetosymmetry, @keefemitman & @moble - The `__mul__` method now uses `multiplication_truncator` to set the `ell_max`. Let me know if its default value - `sum` is a reasonable choice when the attribute isn't set!

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,42 @@ def __getitem__(self, key):
...
         return obj
 
+    def __mul__(self, other):
+        if isinstance(other, WaveformModes) and np.allclose(self.frame, other.frame):
+            if self.multiplication_truncator is not None:
+                new_ell_max = self.multiplication_truncator((self.ell_max, other.ell_max))
+            else:
+                new_ell_max = max(
+                    truncator((self.ell_max, other.ell_max))
+                    for truncator in [
```
keefemitman: Why do this loop here? Why not just use `sum` since you've already checked for `multiplication_truncator` in the lines above?

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,42 @@ def __getitem__(self, key):
                 obj = TimeSeries(obj)
         return obj
 
+    def __mul__(self, other):
+        if isinstance(other, WaveformModes) and np.allclose(self.frame, other.frame):
+            if self.multiplication_truncator is not None:
```
keefemitman: Do you want to also check if `other.multiplication_truncator` is `None`?

keefemitman (changes requested):

sxs/waveforms/waveform_modes.py
akhairna: If their `multiplication_truncator` is `max` for both, then it should return the max ell. Is this line of thought not correct? If one has `max` and other has `sum`, the it would return sum of ells.

sxs/waveforms/waveform_modes.py
akhairna: Yes, I should've added that check too.

sxs/waveforms/waveform_modes.py
akhairna: Oh I think I misunderstood what you're saying. Yes, since I checked I can just do the sum.
Maybe I should move this logic as the default, and do the sum if these attributes are None.

duetosymmetry:
I think there needs to be some clarification on how the two waveforms' multiplication_truncators are used. It needs to be well-documented what happens if they have two different values. I think the simplest would be to use the multiplication_truncator from the left argument, unless it's missing, in which case to use the one from the right argument; unless they are both missing, in which case to fall back on a default.

moble:
This might be a good opportunity to close #21.  Specifically, the [`_ModesTimeSeries`](https://github.com/sxs-collaboration/sxs/blob/0a17c76528c49d522a2c372cc07d2067933ce1b2/sxs/waveforms/waveform_mts.py#L4) subclass is used for a lot of stuff in [memory.py](https://github.com/sxs-collaboration/sxs/blob/0a17c76528c49d522a2c372cc07d2067933ce1b2/sxs/waveforms/memory.py#L1) which could maybe be simplified.  We now have [a good few](https://github.com/sxs-collaboration/sxs/blob/0a17c76528c49d522a2c372cc07d2067933ce1b2/sxs/waveforms/waveform_modes.py#L272-L400) of those features.  It looks like add, subtract, and divide would be the other missing items.

One thing all four of these share is that we need to check that they are comparable.  Your current check that the frames are close is a start, but I think it would be better to have a helper method to check that the frame types are the same, and that the frame and time arrays are not only close, but actually equal.  I don't see how two waveforms would get similar but not equal frames or times.

It looks like these algebraic functions work currently, but just because of the inheritance from np.array, which should be considered a bug; when they involve two WaveformModes, they should always be checked for comparable data.

akhairna:
Hi @moble, I am sorry I had to rebase it on top of the recent commits that were merged on upstream. Otherwise the git history wasn't clean.

akhairna:
Hi @moble, @duetosymmetry, @keefemitman - I incorporated some of the suggestions.
- Added some documentation to the `__mul__` method describing the order in which truncator is used, and its default if missing.
- Addition of method to check compatibility of WModes before multiplication.

Any suggestions on the updated structure of the `__mul__` method and the helper function? I will look into the other algebraic operations too.
Q: After multiplication, I have put the `multiplication_truncator` attribute of the result to be `max`. Do you all think it should be `sum` instead?

duetosymmetry:
Just so we don't forget: this PR is dependent on moble/spherical#28 . After that is merged and spherical has a version bump, the dependencies in pyproject.toml will need to be bumped in this PR.

duetosymmetry:
New dep is `spherical>=1.1.0`

akhairna:
Hi @moble, @keefemitman, and @duetosymmetry. 
Here is an issue of shape of the data array that I am facing. Suppose I want to check this bondi constaint:
`\dot{ψ₀} = ðψ₁ + 3 σ ψ₂`
`σ ψ₂` - I have taken the `abs((σ ψ₂).spin_weight)` of this product to be the `l_min` for slicing the data array. 
`ðψ₁` - Raises the spin but doesn't slice or discard the `l=1` modes.
Operations like these result in mismatch between the shape of the data array which gives an error from `__add__` method. What is a reasonable approach to deal with this?
1. Set the `l_min` to be 0 for all (which we were doing in scri).
2. Allow operations like `eth` and `ethbar` to change the shape of the data array (reasonable according to me).
3. Any other way to handle this?
Please leave any feedback on the existing commits.

akhairna:
@moble, @duetosymmetry & @keefemitman  - Based on the discussion during vacuum call, I have modified the `__add__` and `__sub__` method. All the algebraic operations are working based on a few checks that I performed. Let me know if you have any feedback.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,154 @@ def __getitem__(self, key):
...
+            time_axis=0,
+            ell_min=ell_min,
+            ell_max=ell_max,
+            modes_axis=1,
+            spin_weight=self.spin_weight,
+            frame=self.frame,
+            frame_type=self.frame_type,
+        )
+
+    def __sub__(self, other):
```
keefemitman, line 214: Can we define a `__neg__` function and then make `__sub__` call `__add__` but with a minus sign? This will help avoid rewriting a large block of identical code. Not crucial, but could be nice.

sxs/waveforms/waveform_modes.py
```diff
@@ -173,6 +321,11 @@ def ell_max(self):
         """Largest value of ℓ stored in the data"""
         return self._metadata["ell_max"]
 
+    @property
+    def multiplication_truncator(self):
+        """Multiplication truncator stored in the data"""
+        return self._metadata["multiplication_truncator"]
```
keefemitman: I believe this can error since `multiplication_truncator` need not be set when creating a `WaveformModes` object, no? Probably need to change this to `self._metadata.get("multiplication_truncator", None)`

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,154 @@ def __getitem__(self, key):
...
+        ell_min = min(self.ell_min, other.ell_min)
+        ell_max = max(self.ell_max, other.ell_max)
+
+        result = np.zeros((self.time.size, spherical.Ysize(ell_min, ell_max)), dtype=self.dtype)
+
+        slice_1 = slice(spherical.Yindex(self.ell_min, -self.ell_min, ell_min), spherical.Yindex(self.ell_max + 1, -(self.ell_max + 1), ell_min),)
+
+        slice_2 = slice(spherical.Yindex(other.ell_min, -other.ell_min, ell_min), spherical.Yindex(other.ell_max + 1, -(other.ell_max + 1), ell_min),)
+
+        result[:, slice_1] += self.data
```
keefemitman: This is annoying, but since `time_axis` and `modes_axis` are free options, this could fail if `time_axis != 0` and `modes_axis != 1`.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,154 @@ def __getitem__(self, key):
...
+        self._is_compatible(other)
+
+        if self.multiplication_truncator is not None:
+            new_ell_max = self.multiplication_truncator((self.ell_max, other.ell_max))
+        elif other.multiplication_truncator is not None:
+            new_ell_max = other.multiplication_truncator((self.ell_max, other.ell_max))
+        else:
+            new_ell_max = self.ell_max + other.ell_max
+
+        modes12_data, modes12_ellmin, modes12_ellmax, modes12_spin = spherical.multiply(
```
keefemitman: I believe this will fail if one doesn't have `time_axis == 0` and `modes_axis == 1`.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,154 @@ def __getitem__(self, key):
...
+            modes_data,
+            time=self.time,
+            time_axis=0,
+            ell_min=ell_min,
+            ell_max=modes12_ellmax,
+            modes_axis=1,
+            spin_weight=modes12_spin,
+            frame=self.frame,
+            frame_type=self.frame_type,
+            multiplication_truncator=self.multiplication_truncator
```
keefemitman: Shouldn't this match the multiplication truncator used above?

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,154 @@ def __getitem__(self, key):
...
+        self.spin_weight,
+        other,
+        other.ell_min,
+        other.ell_max,
+        other.spin_weight,
+        ellmax_fg=new_ell_max
+        )
+
+        ell_min = abs(modes12_spin)
+        modes_data = modes12_data[:,spherical.Yindex(ell_min, - ell_min, modes12_ellmin) : ]
```
keefemitman: This could cause issues if `ell_min` ends up larger than `ellmax_fg`. Maybe add a check?

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,154 @@ def __getitem__(self, key):
...
+
+        return WaveformModes(
```
keefemitman: This should be `type(self)`, no?

keefemitman (changes requested):
Thanks for doing this @akhairna. Unfortunately I think there's a few annoying bugs/issues present still.

sxs/waveforms/waveform_modes.py
akhairna: I ran into this bug after pushing these changes. I think we need to think about a default for `multiplication_truncator`. How about `max` rather than `None`?

sxs/waveforms/waveform_modes.py
akhairna: Yes, I should edit it to `type(self)`.

sxs/waveforms/waveform_modes.py
keefemitman: That works, but then you'll never need the call `new_ell_max = self.ell_max + other.ell_max`

akhairna: Yes, I realised it after posting that comment (and I didn't see your message). I will keep it to be `None`.

sxs/waveforms/waveform_modes.py
akhairna: Yes, that would definitely make the code cleaner. Thanks for the suggestion. I will do that.

sxs/waveforms/waveform_modes.py
akhairna: I had discussed this with @duetosymmetry yesterday. We were wondering if there happens to be a case when the axes don't match for the two WModes. Well! we didn't come to a final resolution on what's the best thing to do.

sxs/waveforms/waveform_modes.py
akhairna: okay.

akhairna:
Hi @moble, @keefemitman, and @duetosymmetry. With these new commits, there shouldn't be any issue regarding shape of two WModes while doing algebra. As long as they are compatible via the `is_compatible` function, they can be added, subtracted or multiplied even if they differ in `time_axis`. 
Please review it, and let me know if it looks alright.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,152 @@ def __getitem__(self, key):
...
+        shape = (n_times, n_modes) if self.time_axis==0 else (n_modes, n_times)
+        result = np.zeros(shape, dtype=self.dtype)
+
+        idx1 = (slice(None), slice_1) if self.time_axis == 0 else (slice_1, slice(None))
+        idx2 = (slice(None), slice_2) if self.time_axis == 0 else (slice_2, slice(None))
+
+        result[idx1] += self.data
+        result[idx2] += other.data
+
+        return type(self)(result, **self._metadata)
```
keefemitman: Don't you want to update the metadata to use the new `ell_min` and `ell_max`?

keefemitman: Maybe do something like:
```
metadata = self._metadata.copy()
metadata.update(
    ell_min=ell_min,
    ell_max=ell_max,
    spin_weight=self.spin_weight,
)
return type(self)(result, **metadata)
```

keefemitman (changes requested):
Sorry to request another change @akhairna. But I approve the PR after this.

keefemitman (approved):

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        result[idx2] += other.data
+
+        metadata = self._metadata.copy()
+        metadata.update(
+            ell_min=ell_min,
+            ell_max=ell_max,
+            spin_weight=self.spin_weight,
+        )
+
+        return type(self)(result, metadata)
```
moble: Shouldn't this unpack metadata, so they'll appear as kwargs, rather than args?

```suggestion
        return type(self)(result, **metadata)
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        algebraic operations."""
+        if self.frame_type!=other.frame_type:
+            raise ValueError(
+                f"Both waveforms must have identical frame types; self has `{self.frame_type=}`, other has `{other.frame_type=}`."
+        )
+        if not np.array_equal(self.frame, other.frame):
+            raise ValueError(f"Both waveforms must have identical frame arrays.")
+        if not np.array_equal(self.time, other.time):
+            raise ValueError(f"Both waveforms must have identical time arrays.")
+
```
moble: Add a check here that their time and modes axes are the same.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+    def __truediv__(self, other):
+        """Division of two WaveformModes object.
+        """
+        if isinstance(other, type(self)):
+            raise ValueError(f"Cannot divide one WaveformModes object by another")
+        else:
+            return super().__truediv__(other)
+
+    def _is_compatible(self,other):
```
moble: It doesn't really make sense to name this `_is_compatible` and return a Bool (that gets ignored) if they are, but then raise errors if not.  But we probably do want to see the reason, and fail when they're not compatible, so I guess rename to `_require_compatibility` and remove the return value.

sxs/waveforms/waveform_modes.py
```diff
@@ -1382,8 +1540,34 @@ def remove_memory(self, integration_start_time):
...
+            multiplication_truncator=self.multiplication_truncator
+        )
+
+    @property
+    def T(self):
+        """Returns the transpose of this waveform modes object.
+
+        Equivalent to calling transpose(). See transpose() for details.
+        """
+        return self.transpose()
```
moble: I don't think we should mess around with trying to transpose for the user.  If they've input things with different shapes, I would expect that to be more likely to be due to a bug.  We should check that the the time and modes axes are equal, and if not, just raise an error.  I don't think anything is intended to say that a WaveformModes is just 2-D.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        """
+        if not isinstance(other, type(self)):
+            return super().__mul__(other)
+
+        self._is_compatible(other)
+
+        time_axis = self.time_axis
+
+        left_WM = self if self.time_axis == 0 else self.T
+        right_WM = other if other.time_axis == 0 else other.T
```
moble: Skip this in favor of the compatibility requirement
```suggestion
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+
+        modes_data = modes12_data[:, start_idx: ]
+
+        result = type(self)(
+                modes_data,
+                time=self.time,
+                time_axis=0,
+                ell_min=ell_min,
+                ell_max=modes12_ellmax,
+                modes_axis=1,
```
moble:
```suggestion
                time_axis=self.time_axis,
                ell_min=ell_min,
                ell_max=modes12_ellmax,
                modes_axis=self.modes_axis,
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+                spin_weight=modes12_spin,
+                frame=self.frame,
+                frame_type=self.frame_type,
+                multiplication_truncator=truncator
+            )
+
+        if time_axis != 0:
+            result = result.T
```
moble:
```suggestion
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        The output `ell_max` is determined by the `multiplication_truncator`
+        attribute. The left operand's truncator takes priority; if absent, the
+        right operand's truncator is used; if both are absent, it defaults to
+        the sum of the two `ell_max`.
+        """
+        if not isinstance(other, type(self)):
+            return super().__mul__(other)
+
+        self._is_compatible(other)
+
```
moble, line 247: `spherical.multiply` requires that the last index should be the modes index.  Check that here, and add a note about this to the docstring.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        )
+
+        ell_min = abs(modes12_spin)
+
+        if ell_min > modes12_ellmax:
+            raise ValueError(f"ell_min ({ell_min}) of the product self*other exceeds ell_max ({modes12_ellmax}), resulting in no valid modes.")
+
+        start_idx = spherical.Yindex(ell_min, - ell_min, modes12_ellmin)
+
+        modes_data = modes12_data[:, start_idx: ]
```
moble: In case there are more than 2 dimensions
```suggestion
        modes_data = modes12_data[..., start_idx: ]
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+
+        self._is_compatible(other)
+
+        time_axis = self.time_axis
+
+        left_WM = self if self.time_axis == 0 else self.T
+        right_WM = other if other.time_axis == 0 else other.T
+
+        truncator = left_WM.multiplication_truncator or right_WM.multiplication_truncator or sum
+        new_ell_max = truncator((left_WM.ell_max, right_WM.ell_max))
```
moble: I thought the conclusion of our discussions was just to use whichever was more conservative.  Maybe something like this:
```suggestion
        self_ell_max = self.multiplication_truncator((self.ell_max, other.ell_max))
        other_ell_max = other.multiplication_truncator((self.ell_max, other.ell_max))
        if self_ell_max >= other_ell_max:
            truncator = self.multiplication_truncator
            new_ell_max = self_ell_max
        else:
        	truncator = other.multiplication_truncator
        	new_ell_max = other_ell_max
        end
```
This would also require a functional default for `WaveformModes.multiplication_truncator`; see below.

sxs/waveforms/waveform_modes.py
```diff
@@ -173,6 +326,11 @@ def ell_max(self):
         """Largest value of ℓ stored in the data"""
         return self._metadata["ell_max"]
 
+    @property
+    def multiplication_truncator(self):
+        """Multiplication truncator stored in the data"""
+        return self._metadata.get("multiplication_truncator", None)
```
moble: I lean towards assuming something that will at least function.  If we're being conservative elsewhere, we can afford to be lenient here:
```suggestion
        return self._metadata.get("multiplication_truncator", max)
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+            right_WM.ell_min,
+            right_WM.ell_max,
+            right_WM.spin_weight,
+            ellmax_fg=new_ell_max
+        )
+
+        ell_min = abs(modes12_spin)
+
+        if ell_min > modes12_ellmax:
+            raise ValueError(f"ell_min ({ell_min}) of the product self*other exceeds ell_max ({modes12_ellmax}), resulting in no valid modes.")
```
moble: Wouldn't it make more sense to check this *before* trying to multiply?  We know what the spin weight will be, and what `new_ell_max` we're requesting.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+
+        modes12_data, modes12_ellmin, modes12_ellmax, modes12_spin = spherical.multiply(
+            left_WM,
+            left_WM.ell_min,
+            left_WM.ell_max,
+            left_WM.spin_weight,
+            right_WM,
+            right_WM.ell_min,
+            right_WM.ell_max,
+            right_WM.spin_weight,
```
moble: The above would eliminate `left_WM` and `right_WM`, so probably just stick with `self` and `other`.

sxs/waveforms/waveform_modes.py
```diff
@@ -173,6 +326,11 @@ def ell_max(self):
         """Largest value of ℓ stored in the data"""
         return self._metadata["ell_max"]
 
+    @property
+    def multiplication_truncator(self):
+        """Multiplication truncator stored in the data"""
+        return self._metadata.get("multiplication_truncator", None)
+
```
moble, line 341: Should probably add a `@multiplication_truncator.setter` function so that people can, e.g., change defaults.

sxs/waveforms/format_handlers/rotating_paired_diff_multishuffle_bzip2.py
```diff
@@ -567,6 +567,7 @@ def invalid(message):
         ell_min=ell_min,
         ell_max=ell_max,
         spin_weight=spin_weight,
+        multiplication_truncator=max
```
moble: I think this should just be a `WaveformModes` default, rather than needing to specify it here.
```suggestion
```

moble (changes requested):

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        )
+
+        if self.time_axis != other.time_axis:
+            other = other.T
+
+        n_modes = spherical.Ysize(ell_min, ell_max)
+        n_times = self.n_times
+
+        shape = (n_times, n_modes) if self.time_axis==0 else (n_modes, n_times)
+        result = np.zeros(shape, dtype=self.dtype)
```
moble: This is a bit more consistent
```suggestion
        result = np.zeros(shape, dtype=np.result_type(self, other))
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+
+        slice_1 = slice(
+            spherical.Yindex(self.ell_min, -self.ell_min, ell_min),
+            spherical.Yindex(self.ell_max + 1, -(self.ell_max + 1), ell_min),
+        )
+
+        slice_2 = slice(
+            spherical.Yindex(other.ell_min, -other.ell_min, ell_min),
+            spherical.Yindex(other.ell_max + 1, -(other.ell_max + 1), ell_min),
+        )
```
moble, line 187:
```suggestion
        # Compute slices into `result` corresponding to `self` and `other`'s modes axes
        slice_1 = slice(
            spherical.Yindex(self.ell_min, -self.ell_min, ell_min),
            spherical.Yindex(self.ell_max, self.ell_max, ell_min) + 1,
        )
        slice_2 = slice(
            spherical.Yindex(other.ell_min, -other.ell_min, ell_min),
            spherical.Yindex(other.ell_max, other.ell_max, ell_min) + 1,
        )
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+
+        if self.time_axis != other.time_axis:
+            other = other.T
+
+        n_modes = spherical.Ysize(ell_min, ell_max)
+        n_times = self.n_times
+
+        shape = (n_times, n_modes) if self.time_axis==0 else (n_modes, n_times)
```
moble: I think something like this should allow broadcasting
```suggestion
        shape = list(np.broadcast_shapes(self.shape, other.shape))
        shape[self.modes_axis] = n_modes
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,159 @@ def __getitem__(self, key):
...
+        result = np.zeros(shape, dtype=self.dtype)
+
+        idx1 = (slice(None), slice_1) if self.time_axis == 0 else (slice_1, slice(None))
+        idx2 = (slice(None), slice_2) if self.time_axis == 0 else (slice_2, slice(None))
```
moble: I think this is the other part needed for broadcasting
```suggestion
        idx1 = [slice(None)] * self.ndim
        idx1[self.modes_axis] = slice_1
        idx2 = [slice(None)] * other.ndim
        idx2[other.modes_axis] = slice_2
```

moble (changes requested):

sxs/waveforms/waveform_modes.py
```diff
@@ -173,6 +332,15 @@ def ell_max(self):
         """Largest value of ℓ stored in the data"""
         return self._metadata["ell_max"]
 
+    @property
+    def multiplication_truncator(self):
+        """Multiplication truncator stored in the data"""
+        return self._metadata.get("multiplication_truncator", max)
+
+    @multiplication_truncator.setter
+    def multiplication_truncation(self, truncator_prm):
```
moble: The names have to match
```suggestion
    def multiplication_truncator(self, truncator_prm):
```

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,165 @@ def __getitem__(self, key):
...
+
+    def __mul__(self, other):
+        """Multiplication method for two WaveformModes object.
+
+        First it checks if the two objects are compatible for multiplication
+        using the `_require_compatibility` method.
+        The output `ell_max` is determined by the `multiplication_truncator`
+        attribute. The left operand's truncator takes priority; if absent, the
+        right operand's truncator is used; if both are absent, it defaults to
+        the sum of the two `ell_max`.
```
moble: This should be updated to reflect what actually happens now.

moble (changes requested):

akhairna:
Hi @moble. I added two tests and made some minor changes.
I changed line 174 in waveform_modes.py to ` n_modes = max(self.n_modes, other.n_modes)`. This relies directly on the shape of the data rather than computing the number of modes between `l_min` and `l_max`. If there are missing modes, this would handle such cases appropriately.

akhairna:
@moble, @duetosymmetry & @keefemitman - Can you all please review these tests and the final changes? If this PR is merged, then I will be able to test my ABD class branch.

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,167 @@ def __getitem__(self, key):
...
+            raise ValueError(
+                f"Cannot add two WaveformModes with different spin weights ({self.spin_weight=} and {other.spin_weight=})."
+            )
+
+        self._require_compatibility(other)
+
+        ell_min = min(self.ell_min, other.ell_min)
+        ell_max = max(self.ell_max, other.ell_max)
+
+        n_modes = max(self.n_modes, other.n_modes)
```
keefemitman: Shouldn't `n_modes` be based on `ell_min` and `ell_max`?

sxs/waveforms/waveform_modes.py
```diff
@@ -150,6 +150,167 @@ def __getitem__(self, key):
...
+        Note that the `modes_axis` of self and other should be the last axis of
+        the WaveformModes data as per spherical.multiply, otherwise a
+        ValueError is raised.
+        """
+        if not isinstance(other, type(self)):
+            return super().__mul__(other)
+
+        self._require_compatibility(other)
+
+        if self.modes_axis != self.ndim - 1:
```
keefemitman, line 243: Should we also require `other.modes_axis == other.ndim - 1`?

keefemitman (changes requested):

keefemitman (approved):

moble:
It's really unhelpful when you force push to a PR.  You'll never *need* to do it, and ***you should never do it***.  Github doesn't try to track through a force push, so it invalidates all the previous review history; we can't see what's new or what happened to our previous comments and suggestions.  Serious bugs have snuck into the code this way in the past, and it makes us have to start basically from scratch in reviewing.  Don't worry about what has happened on main, or a messy commit history; I can deal with the squash and merge.  If you need some new bug fix from main, just make a merge commit (or cherry pick) rather than rebasing; again, it will be fine on squash and merge.  Please never do this again on a PR that I have to look at.

akhairna:
Hi @moble, can you please leave comments or feedback that you mentioned during the vacuum call?

PR #203, opened 2026-06-29
Add RITSimulation_v5 to load waveforms from v5 of the catalog.

akhairna:
Hi @moble & @duetosymmetry - I am adding this subclass `RITSimulation_v5` to load waveforms from the latest RIT catalog release. I needed some suggestions on the structure:

In the function `RITSimulation`, I am forcing the version to be `v5` so that it uses just the RITSimulation_v5 class. Here `v5` corresponds to the catalog version rather than version of the waveforms. I think a better structure would be to merge `RITSimulation_v4` and `RITSimulation_v5` class into a single class called `RITSimulationBase`. If there are multiple versions of the simulations in the future, then we can create a different version using base class. 

I agree it would a significant change. I should have thought of organizing this the first time. Let me know what would be appropriate, and any feedback on the code.

----
📚 Documentation preview 📚: https://sxs--203.org.readthedocs.build/en/203/

### DIFF
--- a/sxs/simulations/rit_simulation.py
+++ b/sxs/simulations/rit_simulation.py
@@ -70,22 +70,25 @@ def RITSimulation(location, *args, **kwargs):
 
     resolution_tags = metadata.get("resolution_tags", [metadata.resolution_tag])
 
-    if resolution_tag is not None and resolution_tag not in resolution_tags:
-        raise ValueError(
-                f"Resolution tag '{resolution_tag}' not found in simulation files for {rit_id}."
-            )
-
     # If res is given as part of `location` use it; otherwise, use the highest
     # available.
     if resolution_tag is None:
         resolution_tag = max(resolution_tags, key = lambda r: int(r[1:]))
 
+    if resolution_tag not in resolution_tags:
+        raise ValueError(
+                f"Resolution tag '{resolution_tag}' not found in simulation files for {rit_id}."
+            )
+
+
     location = f"{rit_id}/{resolution_tag}"
 
+    # This adds "files" key to the metadata for backwards compatibility with v4
+    # tag of RITSimulations.json.
     if "files" not in metadata:
         files = {
         f"{resolution_tag}:extrap_strain.h5 ": {"link": metadata.extrap_strain_url},
-        f"{resolution_tag}:extrap_psi4.h5": {"link": metadata.extrap_psi4_url},
+        f"{resolution_tag}:extrap_psi4.tar.gz": {"link": metadata.extrap_psi4_url},
         f"{resolution_tag}:metadata.txt": {"link": metadata.metadata_url},
     }
         metadata.update(files=files)
--- a/sxs/simulations/rit_simulation.py
+++ b/sxs/simulations/rit_simulation.py
@@ -35,7 +35,7 @@ def RITSimulation(location, *args, **kwargs):
     "RIT:eBBH:xxxx". It must exist, or no files will be found to load the data
     from.
 
-    The returned object will be a `RITSimulation_v4` object.
+    The returned object will be a `RITSimulation_v5` object.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


duetosymmetry:
This depends on sxs-collaboration/sxscatalog#18 which should be merged first

sxs/simulations/rit_simulation.py
```diff
@@ -48,29 +57,56 @@ def RITSimulation(location, *args, **kwargs):
...
-    series = simulations.dataframe.loc[location]
-    metadata = Metadata(dict(series))
-    files = {
-        "strain": {"link": metadata.extrap_strain_url},
-        "psi4": {"link": metadata.extrap_psi4_url},
-        "metadata": {"link": metadata.metadata_url},
+    series = simulations.dataframe.loc[rit_id]
+    metadata = simulations[rit_id]
+
+    resolution_tags = metadata.resolution_tags if "resolution_tags" in metadata else [metadata.resolution_tag]
```
duetosymmetry:
```suggestion
    resolution_tags = metadata.get("resolution_tags", [metadata.resolution_tag])
```

sxs/simulations/rit_simulation.py
```diff
@@ -48,29 +57,56 @@ def RITSimulation(location, *args, **kwargs):
...
+
+    if resolution_tag is not None and resolution_tag not in resolution_tags:
+        raise ValueError(
+                f"Resolution tag '{resolution_tag}' not found in simulation files for {rit_id}."
+            )
+
+    # If res is given as part of `location` use it; otherwise, use the highest
+    # available.
+    if resolution_tag is None:
+        resolution_tag = max(resolution_tags, key = lambda r: int(r[1:]))
```
duetosymmetry: If you move this above the preceding check, then you can skip the part `resolution_tag is not None `

sxs/simulations/rit_simulation.py
```diff
@@ -48,29 +57,56 @@ def RITSimulation(location, *args, **kwargs):
...
+
+    location = f"{rit_id}/{resolution_tag}"
+
+    if "files" not in metadata:
```
duetosymmetry: Leave a comment that this is for backwards compatibility with the v4 metadata JSON that we generated

sxs/simulations/rit_simulation.py
```diff
@@ -48,29 +57,56 @@ def RITSimulation(location, *args, **kwargs):
...
+        files = {
+        f"{resolution_tag}:extrap_strain.h5 ": {"link": metadata.extrap_strain_url},
+        f"{resolution_tag}:extrap_psi4.h5": {"link": metadata.extrap_psi4_url},
```
duetosymmetry: Here the extension ought to match (.tar.gz)

sxs/simulations/rit_simulation.py
```diff
@@ -48,29 +57,56 @@ def RITSimulation(location, *args, **kwargs):
...
+        f"{resolution_tag}:metadata.txt": {"link": metadata.metadata_url},
     }
-    version = "v4.0"
+        metadata.update(files=files)
+
+    files = {k: v for k,v in metadata.files.items() if k.startswith(f"{resolution_tag}")}
```
duetosymmetry:
```suggestion
    files = {k: v for k,v in metadata.files.items() if k.startswith(resolution_tag)}
```

akhairna:
Hi @duetosymmetry and @moble - I removed a few lines from the function "RITSimulation" that were supposed to be for backwards compatibility. Even with an older version of sxs, the function would still download the latest `RITSimulations.json` file. The newer json file will always have the key "files", but it would fail for different reason. The newer json file has renamed the previous keys namely `extrap_strain_url`, `metadata_url` etc and stored them into a dict, due to multiple url from different resolutions of the same simulation. If the user updates their sxs version, then this issue won't occur. Let me know what you think.

I have updated the tests as well so they shouldn't fail.

PR #204, opened 2026-06-30
Modify align*d arg-handling to deal with new scipy failures to pickle

moble:
----
📚 Documentation preview 📚: https://sxs--204.org.readthedocs.build/en/204/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,12 @@ concurrency:
 env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
+  # Force uv (and hatch, which installs via uv here) to use managed
+  # python-build-standalone interpreters rather than any distro Python.
+  # Ubuntu 24.04 ships a system python3.12 linked against OpenSSL 3.0, but
+  # juliapkg refuses Julia >=1.12 unless the running Python's OpenSSL is >=3.5
+  # (Julia 1.12 requires OpenSSL 3.5). Managed builds ship OpenSSL >=3.5.
+  UV_PYTHON_PREFERENCE: "only-managed"
 
 jobs:
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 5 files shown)


moble:
Fixing new multiprocessing pickling failure in waveform alignment

### Problem
`align2d`/`align4d` failed under multiprocessing with:

    TypeError: cannot pickle 'module' object

surfacing in CI on the newest Python. Root cause: scipy 1.18 (Array API adoption)
caches the array-namespace *module* on each `CubicSpline` instance (`_xp`,
`_xp_internal`), and module objects aren't picklable. The brute-force search sent
the cost function into `Pool.map` as a `functools.partial(...)` that captured those
splines, so pickling choked. This worked before only because Linux defaulted to the
`fork` start method (no pickling); Python 3.14 changes the default to `spawn`
everywhere, which forces pickling and exposes the bug.

### Fix
Refactored the multiprocessing in both `align2d` and `align4d`:

- Dropped the `partial(...)` + `pool.map(...)` pattern.
- A `CubicSpline` is a `PPoly`; its `.c`/`.x` arrays are plain, picklable numpy
  arrays. We send those to the workers and reconstruct the spline with
  `PPoly.construct_fast(...)` (no copy, no solve) via small top-level helpers
  (`_picklable_args` / `_reconstruct_args` / `_worker_cost2d` / `_worker_cost4d`),
  feeding args alongside each candidate with `starmap(zip(..., repeat(args)))`.
- Also simplified the serial/optimization path to use `least_squares(..., args=(args,))`,
  removing the per-call `cost_wrapper` closures.

This works under **any** start method (`spawn`/`fork`/`forkserver`) with no
platform-specific handling or module-level global state. Numerical behavior,
return values, interfaces, and the `nprocs == -1` serial path are unchanged.

### Results are bit-for-bit identical
`CubicSpline.__call__` *is* `PPoly.__call__` (no subclass override), and we pass the
already-solved coefficient arrays through unchanged, so there's no numerical
operation that can diverge. Verified across 200 randomized trials (complex-valued
modes, random knot spacings; points in the interior, exactly at knots, and
extrapolated) — every output passes `np.array_equal`, worst diff `0.0`. No
truncation-level drift, so this is a normal (non-major) release.

### CI
Updated the test matrix in `.github/workflows/build.yml`: dropped `3.10`, now tests
`3.12`, `3.13`, `3.14`.

PR #205, opened 2026-07-20
Add standard fields to metadata of RIT simulation

akhairna:
Hi @duetosymmetry and @moble - This PR adds a few extra fields to the RIT metadata. These fields exist in the dataframe, but are absent when the metadata is loaded from the `RITsimulations.json` file or `metadata.txt` file.

----
📚 Documentation preview 📚: https://sxs--205.org.readthedocs.build/en/205/


duetosymmetry:
I would try to avoid having hardcoded field names both here and in sxscatalog ... that makes it hard to remember to update both if you ever change one. So I would recommend having a function in sxscatalog that can update either a `Metadata` object or pandas dataframe with all the fields, and so avoid duplicating this code and the code at https://github.com/sxs-collaboration/sxscatalog/blob/0ce1a8c6706fc61f58795fe1155f0143111707ef/src/sxscatalog/simulations/rit_maya_simulations.py#L288-L301 which is very similar

akhairna:
Hi @moble and @duetosymmetry - This PR is now dependent on sxs-collaboration/sxscatalog#19. Please let me know if there are any suggestions or feedback.

duetosymmetry:
Looks good to me! I guess we will need to wait for a new version of sxscatalog to make sure we're getting the right version of `_add_parameters_to_RIT`

duetosymmetry (approved):
LGTM

akhairna:
@moble - This PR is ready to be merged.

PR #207, opened 2026-08-25
Adding eth and ethbar property to maintain the shape

akhairna:
Hi @moble, @duetosymmetry, and @keefemitman 

I added the method `eth` and `ethbar` to `WaveformModes` object so that it adjusts the shape of the resulting operation to start from `ell_min` = `abs(eth_self.spin_weight)`. The spherical package was setting the shape of the result to be `metadata['ell_min'] = min(abs(self.spin_weight+1), self.ell_min)` (for eth & analogously for ethbar).

I encountered this issue of shape while using the `remove_memory` method of WModes. The function internally calls these operators, and it adds `l=0` and `l=1` modes (which are zero) to the data. I think it is better to not change the shape of the data after removing memory. Let me know if you have any suggestions or feedback on these changes.

----
📚 Documentation preview 📚: https://sxs--207.org.readthedocs.build/en/207/

### DIFF
--- a/sxs/waveforms/waveform_modes.py
+++ b/sxs/waveforms/waveform_modes.py
@@ -572,8 +572,61 @@ class WaveformModes(WaveformMixin, TimeSeries):
     from spherical.modes.derivatives import (
         Lsquared, Lz, Lplus, Lminus,
         Rsquared, Rz, Rplus, Rminus,
-        eth, ethbar
     )
+
+    @property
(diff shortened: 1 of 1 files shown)


akhairna:
Hi @moble, if you find time can you please review this PR?

akhairna:
@duetosymmetry - do you have any comments for this PR? Can you approve it, if it seems alright?

PR #208, opened 2026-09-09
Handle RPDMB with metadata stored inside the H5 file

moble:
We'll soon be distributing RPDMB waveforms with their metadata stored inside the H5 file to make @markscheel happy.

----
📚 Documentation preview 📚: https://sxs--208.org.readthedocs.build/en/208/


moble:
Test failures are just due to Zenodo being down.

PR #209, opened 2026-09-21
Pass tag kwarg to the function Simulation

akhairna:
Hi @moble and @duetosymmetry - This PR will allow the kwarg `tag` to be used while loading an SXS or RIT simulation. This will prevent the call to `get_latest_release` method in sxscatalog.Simulations class (https://github.com/sxs-collaboration/sxscatalog/blob/ac522b3ac2a50716b3007ab80f51e09cac6938d8/src/sxscatalog/simulations/simulations.py#L206) when the tag is passed. 

I was getting `rate_limit_exceeded` type of error when performing multiple calls to `sxs.load`.
```raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/sxs-collaboration/sxscatalogdata/releases
```

This change should avoid this. Please review the PR and let me know if it is an appropriate change.

----
📚 Documentation preview 📚: https://sxs--209.org.readthedocs.build/en/209/


duetosymmetry:
This looks fine to me. @moble, any thoughts? Should we have a list of all kwargs that are allowed to be forwarded?

Issue #110, opened 2024-09-24
Provide (interactive) documentation for the meaning of fields in metadata


duetosymmetry:

duetosymmetry:
@moble do you have a recommendation for where in the package the field documentation should live?

duetosymmetry:
Partially addressed in sxs-collaboration/papers-202X-catalog-update#7, some of that code can be pulled in here once we're happy with it

moble:
I suppose `sxs/sxs/metadata/metadata_fields.py` would be the place for it, right?

duetosymmetry:
I named it `metadata_docs.py`, but I don't strongly prefer either one. Do you have suggestions about the class structure, and whether to provide a list or dict or both?

Also, we will have to do a bit of archaeology to figure out which fields existed at different versions.

duetosymmetry:
Ping, we should do this!

duetosymmetry:
Started in #210

PR #210, opened 2026-09-21
Document what our metadata fields are supposed to mean

duetosymmetry:
This first stab is just pulling in the metadata field docs from when writing the catalog paper. Partially addresses #110 .

Arguably for reflecting in python, we don't want the `description` to be LaTeX, but rather markdown or RST or whatever. Then we would be able to programmatically generate a table to go into the docs.

It's easy enough for the user to just look up a field in a dict; I don't think a `metadata_doc(field)` function is needed.

At a later time, we probably want some sort of checks that every field that exists in the data is actually documented.

----
📚 Documentation preview 📚: https://sxs--210.org.readthedocs.build/en/210/


sxs/metadata/metadata_docs.py
```diff
@@ -0,0 +1,779 @@
...
+    "inputParam": "Input parameters for initial data",
+    "IDmeasure": "Measurements of initial data",
+    "referenceParam": "Reference quantities",
+    "finalProperties": "Properties of merger/final quantities",
+    "codeInfo": "Code information",
+    "timeStamps": "Time stamps information",
+}
+
+def tex_escape(s: str) -> str:
+    return s.replace('_', r'\_')
```
nilsdeppe: Are we worried about a field already have `\_` and this then becoming `\\_`?

nilsdeppe (approved):

